PR #32, opened 2019-10-24
Bump to cbmc-develop and make compileable

xbauch:
One commit on top of: https://github.com/diffblue/hw-cbmc/tree/cbmc-develop

The verilog regression tests are passing.

1: there is no `bmct` so hw-cbmc uses the verifier factory (but somehow wrongly, see below)
2: `aigt` support now lives in trans-netlist (copied from 5.8)

TODO:
1: hw-cbmc 
2: stm2 support (smt decision procedure don't have access to `propt`, have to find out how to convert properties at world-level)


src/hw-cbmc/hw_cbmc_parse_options.cpp
```diff
@@ -50,66 +59,80 @@ int hw_cbmc_parse_optionst::doit()
...
+  unwind_no_timeframes = get_bound();
+  unwind_module = get_top_module();
+  do_unwind_module(prop_conv);
+
+  // the 'extra constraints'
+  if (!constraints.empty()) {
+    log.status() << "converting constraints" << messaget::eom;
+
+    for (const auto &constraint : constraints) {
+      prop_conv.set_to_true(constraint);
```
peterschrammel: I don't think this is doing anything because this is a different solver than used by the verifier. You'd need to write an extension of the goto verifier and checker that allows you to pass in extra constraints.

src/trans-word-level/show_modules.cpp
```diff
@@ -76,7 +76,11 @@ void show_modules(
         std::cout << "  Name:       " << symbol.display_name() << std::endl
                   << std::endl;
         break;
-      
+
+      case ui_message_handlert::uit::JSON_UI:
+        assert(false);
```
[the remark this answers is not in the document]
xbauch: Done.

src/trans-word-level/counterexample_word_level.cpp
```diff
@@ -37,7 +39,9 @@ void show_state(
 
   const symbol_tablet &symbol_table=ns.get_symbol_table();
   
-  forall_symbol_module_map(it, symbol_table.symbol_module_map, module)
+  for(auto it=symbol_table.symbol_module_map.lower_bound(module);
```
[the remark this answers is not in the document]
xbauch: Since we only want to visit the symbols with `module` we propably have to get the range using `upper` and `lower_bound`. And these return iterators without any simple way of getting the `begin` `end` functionality needed for ranged for loop.

src/hw-cbmc/hw_cbmc_parse_options.cpp
xbauch: I commented out the extraction of `prop_conv` and added some comments as to what needs to be done next to get this working.

src/hw-cbmc/hw_cbmc_parse_options.cpp
```diff
@@ -84,11 +84,8 @@ int hw_cbmc_parse_optionst::doit()
       all_properties_verifier_with_trace_storaget<multi_path_symex_checkert>>(
       options, ui_message_handler, goto_model);
 
-  solver_factoryt sf{options, namespacet{goto_model.symbol_table},
-                     ui_message_handler, false};
-  auto my_solver = sf.get_solver();
-  prop_convt &prop_conv =
-      static_cast<prop_convt &>(*my_solver->decision_procedure_ptr);
+  // TODO : extend the interface of verifier to expose the prop_convt
```
peterschrammel: Instead of doing do_unwind_module and add_constraints here, they should probably be done inside an implementatipon of goto_verifier/checker.

xbauch: Yes, that sound more elegant. I've re-worded the todo description.

PR #39, opened 2023-09-10
compile with clang 14

kroening:
This enables compilation with clang 14, mostly by disabling warnings.


tautschnig (approved):

tautschnig:
We should soon follow up with a fix to re-enable those warnings.

kroening:
Of course; I'll add CI before that though.

PR #49, opened 2023-09-14
introduce to_integer_non_constant

kroening:


src/util/ebmc_util.h
```diff
@@ -48,7 +48,8 @@ inline unsignedbv_typet uint64_type() {
   return result;
 }
 
-inline bool to_integer(const exprt &expr, mp_integer &int_value) {
+inline bool to_integer_non_constant(const exprt &expr, mp_integer &int_value)
+{
```
tautschnig, line 52: Can you please add a deprecation marker to this one? It seems the failure more of this function primarily is the situation that `expr` is not a constant, and that should really be made explicit at the call sites. The only other ways `to_integer` itself would fail is when `expr` is not of a numeric type, and this is likely a bug in the code base that ought to be caught via invariants.

kroening: done

tautschnig (approved):

tautschnig:
This now needs a rebase to resolve a conflict.

PR #57, opened 2023-09-16
bump cbmc to 5.91.0

kroening:
This bumps the CBMC dependency to 8c6f963e6f09165be4e4c42f1cf2b76c01f3dd65, which is a commit in the develop branch with version 5.91.0.

Furthermore, this PR includes
* the necessary changes to enable compilation and passing the existing tests
* a bump of the EBMC version number to 4.5
* changes to enable CI with clang++


src/verilog/verilog_expr.h
```diff
@@ -18,7 +18,7 @@ Author: Daniel Kroening, kroening@kroening.com
 class hierarchical_identifier_exprt : public binary_exprt
 {
 public:
-  hierarchical_identifier_exprt() : binary_exprt(ID_hierarchical_identifier)
+  hierarchical_identifier_exprt() : binary_exprt(exprt(), ID_hierarchical_identifier, exprt())
```
tautschnig, line 21: This doesn't look like a desirable way to construct expressions?

src/verilog/verilog_expr.h
```diff
@@ -42,7 +42,7 @@ extern inline hierarchical_identifier_exprt
 class function_call_exprt : public binary_exprt
 {
 public:
-  function_call_exprt() : binary_exprt(ID_function_call)
+  function_call_exprt() : binary_exprt(exprt(), ID_function_call, exprt())
```
tautschnig, line 45: Same here: this constructor really shouldn't be used, I think we need one that actually takes these expressions.

src/ebmc/ebmc_base.cpp
```diff
@@ -126,11 +126,11 @@ int ebmc_baset::finish_bmc(prop_conv_solvert &solver) {
...
     status() << "Checking " << property.name << eom;
-    
-    or_exprt or_expr;
-    
+
+    or_exprt or_expr({});
+
     for(auto l : property.timeframe_literals)
-      or_expr.operands().push_back(literal_exprt(!l));
+      or_expr.add_to_operands(literal_exprt(!l));
```
tautschnig, line 133: Can we instead construct exprt::operandst here, `reserve` the right amount of space, and only thereafter create the `or_exprt`?

src/hw-cbmc/next_timeframe.cpp
```diff
@@ -79,14 +79,14 @@ void add_next_timeframe(
...
+    const code_frontend_assignt member_assignment(member_expr1, member_expr2);
     block.add(member_assignment);
   }
 
   // add code to symbol
   symbol.value=block;
 
   // hide and inline
-  symbol.type.set(ID_C_hide, true);
+  //symbol.type.set(ID_C_hide, true);
```
tautschnig, line 90: Doesn't seem to fit the commit message?

src/hw-cbmc/set_inputs.cpp
```diff
@@ -71,6 +71,6 @@ void add_set_inputs(
   symbol.value=block;
 
   // hide and inline
-  symbol.type.set(ID_C_hide, true);
+  // symbol.type.set(ID_C_hide, true);
```
tautschnig, line 74: This was removed in diffblue/cbmc@4e5bf6ea2bd03, and we might want to replace this code rather than just comment it out?

tautschnig (approved):
Approving, but I would appreciate if my comment could be addressed (which may happen in a separate PR).

src/verilog/verilog_expr.h
kroening: I'll simply remove these for now.

src/verilog/verilog_expr.h
kroening: Also removed.

src/ebmc/ebmc_base.cpp
kroening: sure, done

PR #63, opened 2023-09-19
use the test.pl script from the CBMC repo

kroening:
use the test.pl script from the CBMC repo instead of a local copy


tautschnig:
Why do we need a copy at all?

tautschnig (approved):

PR #70, opened 2023-09-23
Verilog: add a class for exceptions in the preprocessor

kroening:
This adds a class for exceptions in the Verilog preprocessor, as opposed to throwing the integer 0.

### DIFF
--- a/src/verilog/verilog_preprocessor.cpp
+++ b/src/verilog/verilog_preprocessor.cpp
@@ -6,11 +6,13 @@ Author: Daniel Kroening, kroening@kroening.com
 
 \*******************************************************************/
 
-#include <fstream>
+#include "verilog_preprocessor.h"
 
 #include <util/config.h>
 
-#include "verilog_preprocessor.h"
+#include "verilog_preprocessor_error.h"
+
+#include <fstream>
 
 /*******************************************************************\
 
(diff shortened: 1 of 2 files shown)


src/verilog/verilog_preprocessor.cpp
```diff
@@ -603,9 +615,7 @@ void verilog_preprocessort::directive()
 
       if(it==defines.end())
       {
-        error().source_location = source_location;
-        error() << "unknown preprocessor directive \"" << text << "\"" << eom;
-        throw 0;
+        throw verilog_preprocessor_errort() << "unknown preprocessor directive \"" << text << "\"";
```
tautschnig, line 618: I'd have a preference towards RAII: could the constructor please require (!) a string to be passed?

tautschnig (approved):
Approving, but I'd appreciate adjusting the interface as suggested below.

src/verilog/verilog_preprocessor.cpp
kroening: The class meets the RAII criteria; in particular, all invariants are satisfied. It's like a container, and as such may be constructed empty.

src/verilog/verilog_preprocessor.cpp
```diff
@@ -591,9 +604,8 @@ void verilog_preprocessort::directive()
 
       if(it==defines.end())
       {
-        error().source_location = source_location;
-        error() << "unknown preprocessor directive \"" << text << "\"" << eom;
-        throw 0;
+        throw verilog_preprocessor_errort()
+          << "unknown preprocessor directive \"" << text << "\"";
```
tautschnig, line 608: I accept that this satisfies RAII, but I am still wondering whether a constructor that takes a `std::string` wouldn't be the better way forward.

tautschnig (approved):

src/verilog/verilog_preprocessor.cpp
kroening: It would be a bit weird, e.g., in this case you would have
`throw verilog_preprocessor_errort("unknown preprocessor directive \"")
          << text << "\"";`

tautschnig: I would have used + instead of << and have passed the entire string to the constructor. But it’s just a matter of taste, no big deal.

kroening: Yes, but then you'll want numbers, tokens (which have a << operator, but no +), etc.

PR #73, opened 2023-09-23
New tokenizer for the Verilog preprocessor

kroening:
This adds a flex-based tokenizer to the Verilog preprocessor.  Note that tokenization for Verilog preprocessing differs from tokenization for parsing Verilog.

### DIFF
--- a/src/verilog/verilog_preprocessor_tokenizer.cpp
+++ b/src/verilog/verilog_preprocessor_tokenizer.cpp
@@ -1,0 +1,110 @@
+/*******************************************************************\
+
+Module: Verilog Preprocessor Tokenizer
+
+Author: Daniel Kroening, kroening@kroening.com
+
+\*******************************************************************/
+
+#include "verilog_preprocessor_tokenizer.h"
+
+#include <util/invariant.h>
+
+void verilog_preprocessor_token_sourcet::skip_until_eol()
+{
+  while(true)
+  {
+    auto token = peek();
+    if(token == '\n' || token.is_eof())
+      break;
+    next_token(); // eat
+  }
+}
+
+void verilog_preprocessor_token_sourcet::skip_ws()
+{
+  while(true)
+  {
+    auto token = peek();
+    if(token != ' ' && token != '\t')
+      break;
+    next_token(); // eat
+  }
+}
+
+std::string
+verilog_preprocessor_token_sourcet::tokent::string_literal_value() const
+{
+  PRECONDITION(kind == STRING_LITERAL);
+  PRECONDITION(text.size() >= 2);
+  std::string result;
+  result.reserve(text.size());
+
+  for(std::size_t i = 1; i < text.size() - 1; i++)
+  {
+    char ch = text[i];
+    result += ch;
+  }
+
+  return result;
+}
+
+bool verilog_preprocessor_token_sourcet::eof()
+{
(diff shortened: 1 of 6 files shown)


src/verilog/verilog_preprocessor_tokenizer.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::string result;
+  result.reserve(text.size());
+
+  for(std::size_t i = 1; i < text.size() - 1; i++)
+  {
+    char ch = text[i];
+    result += ch;
+  }
+
+  return result;
```
tautschnig, line 49: How about `return text.substr(1, text.size() - 2));`

src/verilog/verilog_preprocessor_tokenizer.cpp
```diff
@@ -0,0 +1,114 @@
...
+}
+
+bool verilog_preprocessor_token_sourcet::eof()
+{
+  return peek().is_eof();
+}
+
+auto verilog_preprocessor_token_sourcet::peek() -> const tokent &
```
tautschnig, line 57: Why this syntax?

src/verilog/verilog_preprocessor_tokenizer.cpp
```diff
@@ -0,0 +1,114 @@
...
+    return token;
+  else
+  {
+    get_token_from_stream();
+    peeked = true;
+    return token;
+  }
+}
+
+auto verilog_preprocessor_token_sourcet::next_token() -> const tokent &
```
tautschnig, line 69: Same here: why is this syntax for return types preferable?

src/verilog/verilog_preprocessor_tokenizer.cpp
```diff
@@ -0,0 +1,114 @@
...
+auto verilog_preprocessor_token_sourcet::peek() -> const tokent &
+{
+  if(peeked)
+    return token;
+  else
+  {
+    get_token_from_stream();
+    peeked = true;
+    return token;
+  }
```
tautschnig, line 66: How about:
```
if(!peeked)
{
  get_token_from_stream();
  peeked = true;
}
return token;
```

src/verilog/verilog_preprocessor_tokenizer.cpp
```diff
@@ -0,0 +1,114 @@
...
+  std::istream &_in)
+  : in(_in)
+{
+}
+
+std::size_t
+verilog_preprocessor_tokenizert::yy_input(char *buffer, std::size_t max_size)
+{
+  std::size_t result;
+  for(result = 0; result < max_size;)
```
tautschnig, line 92: How about:
```
std::size_t result = 0;
while(result < max_size)
```

tautschnig (approved):

src/verilog/verilog_preprocessor_tokenizer.cpp
kroening: Apologies, this is missing a bit. I'll need to add un-escaping for `\t`, `\n`, etc.

src/verilog/verilog_preprocessor_tokenizer.cpp
kroening: Changed to an idiomatic `for(result = 0; result < max_size; result++)`

PR #75, opened 2023-09-25
Verilog preprocessor define arguments

kroening:
This implements Verilog 2001 preprocessor defines with parameters.

### DIFF
--- a/src/verilog/verilog_preprocessor.cpp
+++ b/src/verilog/verilog_preprocessor.cpp
@@ -18,7 +18,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 /*******************************************************************\
 
-Function: verilog_preprocessort::filet::make_source_location
+Function: verilog_preprocessort::contextt::make_source_location
 
   Inputs:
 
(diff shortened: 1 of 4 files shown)


src/verilog/verilog_preprocessor.cpp
```diff
@@ -155,43 +214,94 @@ Function: verilog_preprocessort::replace_macros
...
+  -> std::map<std::string, std::vector<tokent>>
+{
+  if(define.parameters.empty())
+    return {};
+
+  if(tokenizer().next_token() != '(')
+    throw verilog_preprocessor_errort() << "expecting define arguments";
+
+  std::vector<std::vector<tokent>> arguments = {{}};
+
```
tautschnig, line 273: Unnecessary right-hand side?

src/verilog/verilog_preprocessor.h
```diff
@@ -28,14 +29,26 @@ class verilog_preprocessort:public preprocessort
...
+  struct definet
+  {
+    using parameterst = std::vector<std::string>;
+    parameterst parameters;
+    std::vector<tokent> tokens;
+  };
+
+  static std::string as_string(const std::vector<tokent> &);
+
+  typedef std::unordered_map<std::string, definet, string_hash> definest;
```
tautschnig, line 43: Use using instead of typedef

src/verilog/verilog_preprocessor.h
```diff
@@ -60,45 +73,84 @@ class verilog_preprocessort:public preprocessort
...
+    {
+    }
 
-    filet(bool _close, std::istream *_in, std::string _filename)
-      : close(_close), in(_in), filename(std::move(_filename)), tokenizer(*in)
+    explicit contextt(const std::vector<tokent> &tokens)
+      : deallocate_in(false),
+        in(nullptr),
+        filename(),
+        tokenizer(new vector_token_sourcet(tokens))
```
tautschnig, line 116: Why does this need to be a pointer?

tautschnig (approved):

src/verilog/verilog_preprocessor.cpp
kroening: Oh, no! The RHS creates an element in the outer vector, which contains an empty vector. This is important.
I'll add a comment.

src/verilog/verilog_preprocessor.h
kroening: sure

src/verilog/verilog_preprocessor.h
kroening: I suppose you mean the `vector_token_sourcet`, not the input stream.
This is the C version of a variant type. One of the constructors makes a `vector_token_sourcet`, the other a `verilog_preprocessor_tokenizert`. With C++17, this could be a `std::variant`.

PR #80, opened 2023-09-26
Verilog: implement defparam

kroening:
This implements support for Verilog's `defparam` module item.


src/verilog/verilog_expr.h
```diff
@@ -14,9 +14,14 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/expr.h>
 
 /// The syntax for these A.B, where A is a module identifier and B
-/// is an identifier within that module.
+/// is an identifier within that module, given als symbol_exprt.
```
tautschnig, line 17: I think this needs some editing. How about: "The syntax of hierarchical Verilog identifiers is A.B, where A is a module identifier B refers to a symbol within that module. B is given as symbol_exprt."

src/verilog/verilog_typecheck.cpp
```diff
@@ -951,6 +955,61 @@ void verilog_typecheckt::convert_function_call_or_task_enable(
...
+
+\*******************************************************************/
+
+void verilog_typecheckt::convert_parameter_override(
+  const verilog_parameter_overridet &module_item)
+{
+  for(auto &assignment : module_item.assignments())
+  {
+    auto lhs = to_binary_expr(assignment).lhs();
+    auto rhs = to_binary_expr(assignment).rhs();
```
tautschnig, line 974: For me, the use of `auto` made things more difficult: I believe this will be `exprt`, and neither a reference nor will it be `const`. I'd suggest to use `const exprt &lhs`, `const exprt &rhs`.

src/verilog/verilog_expr.h
```diff
@@ -14,9 +14,14 @@ Author: Daniel Kroening, kroening@kroening.com
...
 class hierarchical_identifier_exprt : public binary_exprt
 {
+public:
+  const symbol_exprt &op1() const
```
tautschnig, line 21: Maybe `op1` isn't a great choice here? How about `inner`? I think the comment in `verilog_typecheckt::convert_parameter_override` that explains what `op1` is confirms that this method should have a different name.

src/verilog/verilog_typecheck.cpp
```diff
@@ -951,6 +955,61 @@ void verilog_typecheckt::convert_function_call_or_task_enable(
...
+      error().source_location = module_item.source_location();
+      error() << "defparam expected to have a hierachical identifier" << eom;
+      throw 0;
+    }
+
+    const auto &hierarchical_identifier = to_hierarchical_identifier_expr(lhs);
+
+    // The op0 of the hierarchical identifier is the module instance,
+    // the op1 is the name of the parameter in the module.
+    if(hierarchical_identifier.op0().id() != ID_symbol)
```
tautschnig, line 992: See above comment: we really shouldn't need comments to explain use of an API that we have control over.

src/verilog/verilog_typecheck.cpp
```diff
@@ -1558,11 +1621,23 @@ void verilog_typecheckt::convert_statements()
...
+  // These must all be done before any module instantiation,
+  // which use the parameters.
+  for(auto &item : verilog_module_expr.module_items())
+  {
+    if(item.id() == ID_parameter_override)
+      convert_parameter_override(to_verilog_parameter_override(item));
+    else if(item.id() == ID_set_genvars)
+      for(auto &sub_item : item.operands())
+        if(sub_item.id() == ID_parameter_override)
+          convert_parameter_override(to_verilog_parameter_override(sub_item));
```
tautschnig, line 1638: Would appreciate braces around the multi-line bodies here.

src/verilog/verilog_typecheck.cpp
kroening: Now exprt, with a comment that a copy is made.

src/verilog/verilog_typecheck.cpp
kroening: I'll use `module` and `item`, which is the terminology in the standard.

src/verilog/verilog_typecheck.cpp
kroening: done

PR #85, opened 2023-09-28
Verilog: convert_const_expression now returns a value

kroening:
This changes `convert_const_expression` to return a value, instead of using a reference parameter.

This also removes one instance of `to_integer_non_constant`.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -1080,9 +1080,7 @@ Function: verilog_typecheck_exprt::convert_const_expression
 
 \*******************************************************************/
 
-void verilog_typecheck_exprt::convert_const_expression(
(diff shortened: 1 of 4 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1094,17 +1092,33 @@ void verilog_typecheck_exprt::convert_const_expression(
...
+
+  if(tmp_constant.is_true())
+    return 1;
+  else if(tmp_constant.is_false())
+    return 0;
+  else
+  {
+    mp_integer value;
+    if(to_integer(tmp_constant, value))
+    {
```
tautschnig, line 1113: I’d prefer numeric_cast over to_integer.

tautschnig (approved):

PR #88, opened 2023-10-01
introduce ID_set_genvars

kroening:
This adds an ID for `"set_genvars"`, instead of using the raw string.


src/verilog/verilog_generate.cpp
```diff
@@ -63,7 +63,7 @@ void verilog_typecheckt::elaborate_generate_item(
   else
   {
     // no need for elaboration
-    verilog_module_itemt tmp("set_genvars");
+    verilog_module_itemt tmp(ID_set_genvars);
     tmp.add_to_operands(statement);
     irept &variables=tmp.add("variables");
```
tautschnig, line 68: For a future PR: this is really asking for a proper API to avoid this iterative construction of an object, and also the use of the string `"variables"`.

tautschnig (approved):

PR #91, opened 2023-10-03
replace direct access/construction to/of ireps

kroening:
This replaces some direct accesses to fields of ireps and low-level construction of ireps by the appropriate accessors and constructors.

### DIFF
--- a/src/trans-netlist/counterexample_netlist.cpp
+++ b/src/trans-netlist/counterexample_netlist.cpp
@@ -72,16 +72,16 @@ void show_state(
          var.type.id()==ID_unsignedbv ||
          var.type.id()==ID_signedbv)
       {
-        exprt value_expr(ID_constant, var.type);
+        exprt value_expr;
 
         if(var.type.id()==ID_range)
         {
           mp_integer i=binary2integer(value, false);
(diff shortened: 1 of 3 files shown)


src/trans-netlist/counterexample_netlist.cpp
```diff
@@ -72,16 +72,16 @@ void show_state(
...
         if(var.type.id()==ID_range)
         {
           mp_integer i=binary2integer(value, false);
           mp_integer from=string2integer(var.type.get_string(ID_from));
-          value_expr.set(ID_value, integer2string(i+from));
+          value_expr = constant_exprt(integer2string(i + from), var.type);
         }
         else
-          value_expr.set(ID_value, value);
+          value_expr = constant_exprt(value, var.type);
```
tautschnig, line 84: How about only updating/constructing the `value` in those branches and only then building a `value_expr` which would then be `constant_exprt value_expr{value, var.type};`?

tautschnig (approved):

src/trans-netlist/counterexample_netlist.cpp
kroening: I am anticipating that this code is buggy, given that we no longer use a binary value. Bugfix to follow.

PR #93, opened 2023-10-04
ebmc: add two tests for liveness properties

kroening:


regression/ebmc/SVA-LTL/lasso1-counterexample.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+lasso1.sv
+--bound 10 --trace
+^EXIT=10$
+^SIGNAL=0$
+^\[top\.property\.p0\] .* FAILURE$
+--
```
tautschnig, line 7: It would be great to have some explanation (after another `--` line) of the problem that renders this test (and also the other one in this PR) a "KNOWNBUG."

regression/ebmc/SVA-LTL/lasso1-small-bound.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+lasso1.sv
+--bound 10
+^EXIT=0$
+^SIGNAL=0$
+^\[top\.property\.p0\] .* SUCCESS$
```
tautschnig, line 6: Why do we expect the property to pass? Both the source code and the other .desc file suggest that verification should fail.

tautschnig (changes requested):

regression/ebmc/SVA-LTL/lasso1-counterexample.desc
kroening: done

regression/ebmc/SVA-LTL/lasso1-small-bound.desc
kroening: explanation added

tautschnig (approved):

PR #94, opened 2023-10-04
ebmc: rename do_bmc and related methods

kroening:
This renames the `do_bmc` and `finish_bmc` methods to make it obvious which one is bit level and which one is word level.

### DIFF
--- a/src/ebmc/ebmc_base.cpp
+++ b/src/ebmc/ebmc_base.cpp
@@ -85,7 +85,7 @@ ebmc_baset::ebmc_baset(const cmdlinet &_cmdline,
 
 /*******************************************************************\
 
-Function: ebmc_baset::finish_bmc
+Function: ebmc_baset::finish_word_level_bmc
 
   Inputs:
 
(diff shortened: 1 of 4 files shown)


src/ebmc/ebmc_base.cpp
```diff
@@ -522,8 +524,8 @@ int ebmc_baset::do_bmc(prop_conv_solvert &solver, bool convert_only) {
         const namespacet ns(symbol_table);
         CHECK_RETURN(trans_expr.has_value());
         ::unwind(*trans_expr, *message_handler, solver, bound+1, ns, true);
-        result=finish_bmc(solver);
-        #endif
+        result=finish_word_level_bmc(solver);
+#endif
```
tautschnig, line 528: I'd suggest to also move the `#if 0` at this point.

tautschnig (approved):

src/ebmc/ebmc_base.cpp
kroening: sure

PR #97, opened 2023-10-04
ebmc: re-enable SMT-LIB

kroening:
This re-enables the support for SMT-LIB2 solvers, which was disabled in f374bd54197a19d117f1258e32c4242109449831.


tautschnig (approved):

tautschnig:
What exactly is broken about those tests when running with Z3?

kroening:
> What exactly is broken about those tests when running with Z3?

No support for the range type.

PR #99, opened 2023-10-05
ebmc: add --show-formula

kroening:
This prints the formula generated by BMC in a syntax designed for readability.

### DIFF
--- a/src/ebmc/show_formula_solver.h
+++ b/src/ebmc/show_formula_solver.h
@@ -1,0 +1,65 @@
+/*******************************************************************\
+
+Module: Show Formula Solver
+
+Author: Daniel Kroening, kroening@kroening.com
+
+\*******************************************************************/
+
+/// \file
+/// Show Formula Solver
+
+#ifndef EBMC_SHOW_FORMULA_SOLVER_H
+#define EBMC_SHOW_FORMULA_SOLVER_H
+
+#include <solvers/stack_decision_procedure.h>
+
+class show_formula_solvert : public stack_decision_proceduret
+{
+public:
+  // console
+  show_formula_solvert();
+
+  // file
+  explicit show_formula_solvert(std::ostream &_out) : out(_out), console(false)
+  {
+  }
+
+  void set_to(const exprt &, bool value) override;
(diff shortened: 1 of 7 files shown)


src/ebmc/show_formula_solver.h
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  std::size_t get_number_of_solver_calls() const override
+  {
+    return 0;
+  }
+
+  ~show_formula_solvert()
+  {
+  }
```
tautschnig, line 47: Not necessary?

src/ebmc/show_formula_solver.h
```diff
@@ -0,0 +1,70 @@
...
+
+  void push(const std::vector<exprt> &) override;
```
tautschnig, line 49: Nit pick: why is this one placed in the implementation file when the below ones are inline?

tautschnig (approved):

src/ebmc/show_formula_solver.h
kroening: To avoid a warning -- even though the object is never deleted via the base type

src/ebmc/show_formula_solver.h
kroening: I am considering to output the `push` calls, which would then require exposing the include for `console.h`.

src/ebmc/show_formula_solver.h
tautschnig: But then it could be `~show_formula_solvert() = default;`?

src/ebmc/show_formula_solver.h
tautschnig: But there's the other variant of `push` that's not in the .cpp file? Just asking for as little surprise as possible.

src/ebmc/show_formula_solver.h
kroening: Sure

src/ebmc/show_formula_solver.h
kroening: Variant moved.

PR #103, opened 2023-10-06
ebmc: create transition_systemt class and member

kroening:
This groups the symbol table, the main symbol pointer and the transition system expression into one type.

### DIFF
--- a/src/ebmc/bdd_engine.cpp
+++ b/src/ebmc/bdd_engine.cpp
@@ -350,8 +350,8 @@ void bdd_enginet::compute_counterexample(
 
   satcheckt solver{*message_handler};
   bmc_map.map_timeframes(netlist, number_of_timeframes, solver);
-  
-  const namespacet ns(symbol_table);
+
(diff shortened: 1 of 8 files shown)


src/ebmc/bdd_engine.cpp
```diff
@@ -163,8 +163,8 @@ int bdd_enginet::operator()()
       std::cout << "Atomic propositions:\n";
       for(const auto & a : atomic_propositions)
       {
-        std::string as_string=
-          from_expr(namespacet(symbol_table), irep_idt(), a.first);
```
tautschnig, line 167: Just because I am reading this for the first time: should likely use `format` instead of `from_expr`: the empty `irep_idt` will make it fall back to the default formatter, which is C.

tautschnig (approved):

src/ebmc/bdd_engine.cpp
kroening: #111

PR #104, opened 2023-10-06
ebmc: rename get_model to get_transition_system

kroening:
The term 'transition system' is more specific than 'model'.

### DIFF
--- a/src/ebmc/bdd_engine.cpp
+++ b/src/ebmc/bdd_engine.cpp
@@ -131,7 +131,7 @@ int bdd_enginet::operator()()
 {
   try
   {
-    int result=get_model();
+    int result = get_transition_system();
     if(result!=-1) return result;
 
     {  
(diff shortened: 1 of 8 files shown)


src/ebmc/bdd_engine.cpp
```diff
@@ -131,7 +131,7 @@ int bdd_enginet::operator()()
 {
   try
   {
-    int result=get_model();
+    int result = get_transition_system();
     if(result!=-1) return result;
```
tautschnig, line 135: Not for this PR, but still: the `int` return type should be replaced by an enum. This is rather difficult to understand, and also seems hard to maintain.

tautschnig (approved):

tautschnig:
Also, this needs a rebase now that #106 is merged (which introduced an additional use of `get_model`).

PR #105, opened 2023-10-07
bugfix: bitvectors are no longer binary

kroening:
This replaces a conversion from bitvector constants to mp_integer that assumes a binary representation to `numeric_cast_v`, which hides the details of the representation.


tautschnig (approved):
Is there any way to test this/demonstrate the fix of a bug?

PR #109, opened 2023-10-07
ebmc: neural liveness

kroening:
This adds an implementation of neural liveness checking, i.e., a
guess-verify loop that uses a given neural engine to guess candidate ranking
functions for a liveness property.

### DIFF
--- a/src/ebmc/neural_liveness.cpp
+++ b/src/ebmc/neural_liveness.cpp
@@ -1,0 +1,234 @@
+/*******************************************************************\
+
+Module: Neural Liveness
+
+Author: Daniel Kroening, dkr@amazon.com
+
+\*******************************************************************/
+
+#include "neural_liveness.h"
+
+#include <util/prefix.h>
+#include <util/run.h>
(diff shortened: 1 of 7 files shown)


src/ebmc/neural_liveness.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+int neural_livenesst::operator()()
+{
+  if(!cmdline.isset("neural-engine"))
+    throw ebmc_errort() << "give a neural engine";
+
+  int result = get_transition_system(
+    cmdline, message.get_message_handler(), transition_system);
+  if(result != -1)
+    return result;
```
tautschnig, line 62: Perhaps use exceptions, as is now done everywhere else?

tautschnig (approved):

src/ebmc/neural_liveness.cpp
kroening: Yes, coming

kroening: #140

PR #112, opened 2023-10-12
ebmc: use exceptions for error reporting

kroening:
This introduces an exception class, `ebmc_errort`, for signalling errors in ebmc, instead of using integer return values.

### DIFF
--- a/src/ebmc/ebmc_parse_options.cpp
+++ b/src/ebmc/ebmc_parse_options.cpp
@@ -6,11 +6,15 @@ Author: Daniel Kroening, kroening@kroening.com
 
 \*******************************************************************/
 
-#include <iostream>
+#include "ebmc_parse_options.h"
+
+#include <util/config.h>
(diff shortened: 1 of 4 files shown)


src/ebmc/ebmc_parse_options.cpp
```diff
@@ -45,191 +47,197 @@ Function: ebmc_parse_optionst::doit
...
 {
-  if (config.set(cmdline)) {
-    usage_error();
-    exit(CPROVER_EXIT_USAGE_ERROR);
-  }
-  register_languages();
-
-  if(cmdline.isset("version"))
+  try
   {
```
tautschnig, line 51: Isn't starting the `try` right here taking things a little too far? There is quite a bit of code that I wouldn't expect to throw.

kroening: moved

tautschnig (approved):

PR #114, opened 2023-10-12
ebmc: re-enable --diatest

kroening:
This re-enables the long lost --diatest command line option.

### DIFF
--- a/src/ebmc/ebmc_parse_options.cpp
+++ b/src/ebmc/ebmc_parse_options.cpp
@@ -11,8 +11,10 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/config.h>
 #include <util/exit_codes.h>
 #include <util/help_formatter.h>
+#include <util/string2int.h>
 
 #include "bdd_engine.h"
+#include "diatest.h"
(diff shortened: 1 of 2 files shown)


src/ebmc/ebmc_parse_options.cpp
```diff
@@ -65,20 +67,18 @@ int ebmc_parse_optionst::doit()
...
-        return 1;
-      }
+        throw ebmc_errort() << "must provide number of state bits";
 
-      diatest(bound, atoi(cmdline.getval("statebits")));
+      if(!cmdline.isset("bound"))
+        throw ebmc_errort() << "must provide bound";
+
+      auto statebits = unsafe_string2size_t(cmdline.get_value("statebits"));
+      auto bound = unsafe_string2size_t(cmdline.get_value("bound"));
```
tautschnig, line 77: I believe these should use `string2optional_size_t` for user-provided input is being parsed. If either returns `nullopt`, an exception can be thrown.

kroening: done

tautschnig (approved):

PR #115, opened 2023-10-12
random traces: use --trace-steps instead of --bound

kroening:
The term 'bound' is uncommon in random simulation.  Use the common term 'trace steps' instead of bound.  Note that this counts transition steps, not transition system states.  A trace with one step has two states.


tautschnig:
Needs a rebase.

kroening:
> Needs a rebase.

done

tautschnig (approved):

PR #119, opened 2023-10-12
ebmc: split out property parsing from get_transition_system

kroening:
A 'transition system' does not include properties, hence we split out the code that discovers the properties into a new function.

### DIFF
--- a/src/ebmc/ebmc_base.cpp
+++ b/src/ebmc/ebmc_base.cpp
@@ -855,26 +855,6 @@ int ebmc_baset::get_transition_system()
     *transition_system.trans_expr = new_trans_expr;
   }
 
-  // Property given on command line?
-  if(cmdline.isset('p'))
-  {
-    // NuSMV also uses -p
-    if(parse_property(cmdline.get_value('p')))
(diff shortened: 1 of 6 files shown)


src/ebmc/ebmc_base.cpp
```diff
@@ -910,6 +890,43 @@ int ebmc_baset::get_transition_system()
...
+
+int ebmc_baset::get_properties()
+{
+  // Property given on command line?
+  if(cmdline.isset('p'))
+  {
+    // NuSMV also uses -p
+    if(parse_property(cmdline.get_value('p')))
+      return 1;
+  }
```
tautschnig, line 911: Is NuSMV the right role model here? Should we perhaps support both `-p` and `--property` (which is what CBMC uses)?

tautschnig (approved):

src/ebmc/ebmc_base.cpp
kroening: `--property` selects a property, from an identifier, as in CBMC. `-p` takes an expression as argument.

PR #120, opened 2023-10-12
ebmc: Extract get_transition_system

kroening:


tautschnig:
Needs a rebase.

kroening:
> Needs a rebase.

done

src/ebmc/transition_system.cpp
```diff
@@ -0,0 +1,246 @@
...
+  {
+    message.error() << "CONVERSION ERROR" << messaget::eom;
+    return 2;
+  }
+
+  if(cmdline.isset("show-modules"))
+  {
+    show_modules(transition_system.symbol_table, message_handler.get_ui());
+    return 0;
+  }
```
tautschnig, line 211: Is this the only reason we need a `ui_message_handlert` here? Perhaps this should be factored out, or else various other places should own a reference to a `ui_message_handlert` instead of a `message_handlert` to avoid the (repeatedly occurring) `static_cast`.

kroening: Yes, exactly working towards getting rid of `ui_message_handlert` -- coming!

tautschnig (approved):
Approving with the assumption that there is a path to removal of the `static_cast`s introduced in this PR.

src/ebmc/transition_system.cpp
kroening: #127

PR #121, opened 2023-10-12
introduce propertiest

kroening:
This introduces a class named `propertiest`, which holds the list of properties for ebmc.

### DIFF
--- a/src/ebmc/ebmc_base.cpp
+++ b/src/ebmc/ebmc_base.cpp
@@ -76,7 +76,7 @@ Function: ebmc_baset::property_requires_lasso_constraints
 
 bool ebmc_baset::property_requires_lasso_constraints() const
 {
-  for(const auto &p : properties)
+  for(const auto &p : properties.properties)
     if(!p.is_disabled() && requires_lasso_constraints(p.expr))
       return true;
 
@@ -99,7 +99,7 @@ void ebmc_baset::word_level_properties(decision_proceduret &solver)
 {
   const namespacet ns(transition_system.symbol_table);
 
-  for(propertyt &property : properties)
+  for(propertyt &property : properties.properties)
   {
     if(property.is_disabled())
       continue;
@@ -144,8 +144,8 @@ int ebmc_baset::finish_word_level_bmc(stack_decision_proceduret &solver)
   auto sat_start_time = std::chrono::steady_clock::now();
   
   // Use assumptions to check the properties separately
-  
-  for(propertyt &property : properties)
+
+  for(propertyt &property : properties.properties)
   {
     if(property.is_disabled())
       continue;
@@ -207,7 +207,7 @@ int ebmc_baset::finish_word_level_bmc(stack_decision_proceduret &solver)
 
   // We return '0' if the property holds,
   // and '10' if it is violated.
-  return property_failure() ? 10 : 0;
+  return properties.property_failure() ? 10 : 0;
 }
 
 /*******************************************************************\
@@ -228,7 +228,7 @@ int ebmc_baset::finish_bit_level_bmc(const bmc_mapt &bmc_map, propt &solver)
 
   message.status() << "Solving with " << solver.solver_text() << messaget::eom;
 
-  for(propertyt &property : properties)
+  for(propertyt &property : properties.properties)
   {
     if(property.is_disabled())
       continue;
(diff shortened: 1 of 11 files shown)


src/ebmc/ebmc_base.cpp
```diff
@@ -346,14 +346,14 @@ bool ebmc_baset::parse_property(
...
-  properties.back().description="command-line assertion";
-  properties.back().name="command-line assertion";
-  
+  properties.properties.push_back(propertyt());
+  properties.properties.back().expr = expr;
+  properties.properties.back().expr_string = expr_as_string;
+  properties.properties.back().mode = transition_system.main_symbol->mode;
+  properties.properties.back().location.make_nil();
+  properties.properties.back().description = "command-line assertion";
+  properties.properties.back().name = "command-line assertion";
```
tautschnig, line 355: For a follow-up PR: `property` really wants a proper constructor to avoid this incremental build.

src/ebmc/ebmc_properties.cpp
```diff
@@ -0,0 +1,106 @@
...
+          properties.properties.back().name = symbol.name;
+        else
+          properties.properties.back().name = symbol.pretty_name;
+
+        properties.properties.back().expr = symbol.value;
+        properties.properties.back().location = symbol.location;
+        properties.properties.back().expr_string = value_as_string;
+        properties.properties.back().mode = symbol.mode;
+        properties.properties.back().description =
+          id2string(symbol.location.get_comment());
```
tautschnig, line 50: As above: we need a proper constructor...

src/ebmc/ebmc_properties.cpp
```diff
@@ -0,0 +1,106 @@
...
+      }
+
+      catch(int)
+      {
+        return true;
+      }
+    }
+  }
+
+  return false;
```
tautschnig, line 72: May I suggest to return `optionalt<embc_propertiest>` instead of a Boolean (which the caller may fail to parse or might wrongly ignore) and the reference argument?

src/ebmc/ebmc_properties.cpp
```diff
@@ -0,0 +1,106 @@
...
+      p.status = propertyt::statust::DISABLED;
+
+    bool found = false;
+
+    for(auto &p : properties)
+      if(p.name == property)
+      {
+        found = true;
+        p.status = propertyt::statust::UNKNOWN;
+        break;
```
tautschnig, line 93: Why not use `return false;` here, which should allow removing `bool found` altogether?

src/ebmc/ebmc_properties.h
```diff
@@ -0,0 +1,100 @@
...
+
+#include "transition_system.h"
+
+class ebmc_propertiest
+{
+public:
+  struct propertyt
+  {
+  public:
+    unsigned number;
```
tautschnig, line 26: Is `unsigned` the right type here, or should it perhaps be `std::size_t`?

tautschnig: Also, add `= 0` to simplify the default constructor.

src/ebmc/ebmc_properties.h
```diff
@@ -0,0 +1,100 @@
...
+    bvt timeframe_literals;             // bit level
+    exprt::operandst timeframe_handles; // word level
+    std::string description;
+    enum class statust
+    {
+      DISABLED,
+      SUCCESS,
+      FAILURE,
+      UNKNOWN
+    } status;
```
tautschnig, line 41: With `= 0` above and `status = UNKNOWN` it should be possible to make the default constructor `= default;`

src/ebmc/ebmc_properties.h
```diff
@@ -0,0 +1,100 @@
...
+    inline void make_unknown()
+    {
+      status = statust::UNKNOWN;
+    }
+
+    inline propertyt() : number(0), status(statust::UNKNOWN)
+    {
+    }
+
+    trans_tracet counterexample;
```
tautschnig, line 77: Why is this class member listed down here when all the others are at the top?

tautschnig (approved):
Needs a rebase.

src/ebmc/ebmc_properties.h
kroening: will change to `size_t`.

PR #123, opened 2023-10-15
ebmc: split up show_trans_trace by format

kroening:
This splits up `show_trans_trace` by format, i.e., console output, XML and JSON, and hence, the existing VCD output is no longer asymmetric.

### DIFF
--- a/src/trans-netlist/trans_trace.cpp
+++ b/src/trans-netlist/trans_trace.cpp
@@ -168,6 +168,32 @@ void show_trans_state(
 
 /*******************************************************************\
 
+Function: show_trans_trace
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+void show_trans_trace(
+  const trans_tracet &trace,
+  messaget &message,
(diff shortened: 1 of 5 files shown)


src/trans-netlist/trans_trace.cpp
```diff
@@ -165,6 +165,30 @@ void show_trans_state(
...
+
+\*******************************************************************/
+
+void show_trans_trace(
+  const trans_tracet &trace,
+  messaget &message,
+  const namespacet &ns,
+  std::ostream &out)
+{
+  auto l = trace.get_min_failing_timeframe().value_or(trace.states.size() - 1);
```
tautschnig, line 184: Is there a possibility of `trace.states.empty()`?

tautschnig (approved):
Needs a rebase, and perhaps one `PRECONDITION`.

src/trans-netlist/trans_trace.cpp
kroening: Shouldn't, but I'll add a `PRECONDITION`

PR #126, opened 2023-10-16
translation of trans_tracet to XML now returns a value

kroening:
We can avoid a copy with rvalues, and hence, can effectively return the XML data structure instead of using a reference parameter.

### DIFF
--- a/src/trans-netlist/trans_trace.cpp
+++ b/src/trans-netlist/trans_trace.cpp
@@ -168,28 +168,25 @@ void show_trans_state(
 
 /*******************************************************************\
 
-Function: convert
+Function: xml
 
   Inputs:
 
  Outputs:
 
- Purpose:
+ Purpose: Transform trans_tracet to XML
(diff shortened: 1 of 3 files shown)


src/trans-netlist/trans_trace.cpp
```diff
@@ -165,28 +165,25 @@ void show_trans_state(
...
+xmlt xml(const trans_tracet &trace, const namespacet &ns)
 {
   auto min_failing_timeframe_opt = trace.get_min_failing_timeframe();
 
   auto last_time_frame =
     min_failing_timeframe_opt.value_or(trace.states.size() - 1);
 
-  dest=xmlt("trans_trace");
-  
+  xmlt dest = xmlt("trans_trace");
```
tautschnig, line 185: Nit pick: `xmlt dest{"trans_trace"};`

tautschnig (approved):

PR #130, opened 2023-10-17
ebmc: add API for ranking function checking

kroening:
This exposes two functions, `parse_ranking_function` and `is_ranking_function`, to enable the use of ranking function checking as a component.

### DIFF
--- a/src/ebmc/ranking_function.cpp
+++ b/src/ebmc/ranking_function.cpp
@@ -26,50 +26,20 @@ Author: Daniel Kroening, dkr@amazon.com
 #include "ebmc_error.h"
 #include "report_results.h"
 
-/*******************************************************************\
-
-   Class: ranking_function_checkt
-
- Purpose:
-
-\*******************************************************************/
-
-class ranking_function_checkt : public ebmc_baset
-{
-public:
-  ranking_function_checkt(
-    const cmdlinet &_cmdline,
-    ui_message_handlert &_ui_message_handler)
-    : ebmc_baset(_cmdline, _ui_message_handler),
-      ns{transition_system.symbol_table}
-  {
-  }
-
-  int operator()();
-
-protected:
-  namespacet ns;
-
-  exprt parse_ranking_function();
-  propertyt &find_property();
-};
-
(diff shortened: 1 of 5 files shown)


src/ebmc/ranking_function.cpp
```diff
@@ -114,51 +85,79 @@ ebmc_baset::propertyt &ranking_function_checkt::find_property()
...
 
   if(exit_code != -1)
     throw ebmc_errort();
 
   CHECK_RETURN(transition_system.trans_expr.has_value());
 
   // parse the ranking function
-  const auto ranking_function = parse_ranking_function();
+  if(!cmdline.isset("ranking-function"))
+    throw ebmc_errort() << "no candidate ranking function given";
```
tautschnig, line 105: Is this really the right place for performing this check? I'd expect an invariant here for whatever code is to invoke `do_ranking_function` must only do so when this argument is present.

src/ebmc/ranking_function.cpp
```diff
@@ -114,51 +85,79 @@ ebmc_baset::propertyt &ranking_function_checkt::find_property()
...
 
   // find the property
-  exit_code = get_properties();
+  ebmc_propertiest properties;
 
-  if(exit_code != -1)
-    return exit_code;
+  if(ebmc_propertiest::from_transition_system(
+       transition_system, properties, message_handler))
+    throw ebmc_errort();
```
tautschnig, line 115: Is an exception (without explanation) the right thing here? Perhaps use `CHECK_RETURN` instead?

tautschnig (approved):

src/ebmc/ranking_function.cpp
kroening: I'll change `ebmc_propertiest::from_transition_system` to throw its own exceptions.

src/ebmc/ranking_function.cpp
kroening: That is `is_ranking_function` -- the code in `do_ranking_function` could move to ebmc_parse_options.cpp.

PR #134, opened 2023-10-26
ebmc: add a solver factory

kroening:


regression/ebmc/smv/smv3.desc
```diff
@@ -1,4 +1,4 @@
-CORE
+CORE broken-smt-backend
```
tautschnig, line 1: Was this always broken, and we just didn't notice? At least it's not clear to me how this PR would newly break the test.

kroening: It was broken before; we didn't notice since k-induction was hard-wired to use SAT/boolbvt.

src/ebmc/k_induction.cpp
```diff
@@ -94,6 +95,8 @@ int k_inductiont::operator()()
     return 1;
   }
 
+  solver_factory = ebmc_solver_factory(cmdline);
```
tautschnig, line 98: It's a bit weird that this member is only (re-)initialised here. I started wondering whether it should be a pointer instead.

tautschnig (approved):

src/ebmc/ebmc_base.cpp
```diff
@@ -743,3 +802,28 @@ int ebmc_baset::do_compute_ct()
...
+\*******************************************************************/
+
+int ebmc_baset::do_word_level_bmc()
+{
+  const namespacet ns(transition_system.symbol_table);
+
+  auto solver_factory = ebmc_solver_factory(cmdline);
+  auto solver = solver_factory(ns, message.get_message_handler());
+
+  bool convert_only = cmdline.isset("smt2") || cmdline.isset("outfile") ||
```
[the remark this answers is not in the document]
kroening: It won't solve with `--smt2`, you'll just get the file. It may be arguable that Z3 is _the_ SMT2 solver, and pick that as the default.

src/ebmc/k_induction.cpp
kroening: The container is effectively a pointer, so I don't think there's any performance problem here. It might be prettier to do the initialisation in the constructor.

src/ebmc/ebmc_base.cpp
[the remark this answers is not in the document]
tautschnig: Hmm, ok, I guess I'm just hoping for some consistency across the cprover tools (CBMC _will_ solve with `--smt2`).

PR #139, opened 2023-10-27
Verilog typechecker refactor

kroening:


src/verilog/verilog_typecheck_base.cpp
```diff
@@ -116,17 +116,11 @@ Function: verilog_typecheck_baset::array_size
...
   mp_integer size;
 
-  if(type.id()!=ID_array)
-  {
-    error() << "array_size expected array type" << eom;
-    throw 0;
-  }
-
-  if(to_integer_non_constant(to_array_type(type).size(), size))
+  if(to_integer_non_constant(type.size(), size))
```
tautschnig, line 123: Perhaps use `numeric_cast<mp_integer>(type.size())` instead?

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -431,22 +434,21 @@ void verilog_typecheck_exprt::convert_system_function(
...
+      return argument;
     }
     else if(argument.type().id()==ID_unsignedbv)
     {
       exprt tmp(ID_typecast, argument.type());
       tmp.type().id(ID_signedbv);
       tmp.add_to_operands(std::move(argument));
       tmp.add_source_location()=expr.source_location();
-      expr.swap(tmp);
+      return tmp;
```
tautschnig, line 445: Should this code be cleaned up as well? Like this:
````c++
typet new_type{argument.type()};
new_type.id(ID_signedbv);
typecast_exprt tmp{std::move(argument), std::move(new_type)};
tmp.add_source_location() = expr.source_location();
return std::move(tmp);
````

tautschnig (approved):

src/verilog/verilog_typecheck_base.cpp
kroening: done

src/verilog/verilog_typecheck_expr.cpp
kroening: done

PR #150, opened 2023-10-31
Verilog: parameter ports

kroening:
This adds support for Verilog 2001 style parameter ports.


tautschnig (approved):
Needs a rebase now that the parser has had its rules re-ordered.

PR #151, opened 2023-10-31
Verilog: sort the grammar according to the standard

kroening:
This sorts the rules in the grammar according to their section number in IEEE Std 1800-2017.


tautschnig (approved):
Approving without claiming that I have done any serious equivalence check.

PR #156, opened 2023-11-03
Verilog: implement $bits

kroening:
This implements the `$bits` system function for expression arguments.  It returns the number of bits required to hold an expression as a bit stream.


kroening:
> Should this one be rebased? Presumably `$bits` is also subject to constant folding?

Yes, done.

PR #157, opened 2023-11-04
Verilog: preserve the location of module instance parameter values

kroening:


src/verilog/verilog_typecheck.cpp
```diff
@@ -551,20 +551,23 @@ void verilog_typecheckt::convert_inst(verilog_instt &inst)
...
+      auto &value = static_cast<exprt &>(it->add(ID_value));
+      mp_integer v_int = convert_const_expression(value);
+      value = from_integer(v_int, integer_typet());
+      value.add_source_location() = source_location;
     }
     else
     {
       mp_integer v_int = convert_const_expression(*it);
       *it=from_integer(v_int, integer_typet());
+      it->add_source_location() = source_location;
```
tautschnig, line 570: Could actually use `.with_source_location<exprt>(*it);`.

tautschnig (approved):
Needs a rebase and could be implemented via `with_source_location`.

PR #162, opened 2023-11-07
CI: add CentOS 8 job

kroening:
This adds a job for building and testing ebmc on CentOS 8, which is the
currently 'active' EDA operating system.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -51,7 +51,7 @@ jobs:
       - name: Print ccache stats
         run: ccache -s
 
-  # This job takes approximately ? minutes
+  # This job takes approximately 15 minutes
   check-ubuntu-20_04-make-clang:
     runs-on: ubuntu-20.04
     env:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-checks.yaml
```diff
@@ -100,3 +100,45 @@ jobs:
...
+    container:
+      image: centos:8
+    steps:
+      - name: Install Packages
+        run: |
+          sed -i -e "s|mirrorlist=|#mirrorlist=|g" -e "s|#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g" /etc/yum.repos.d/CentOS-Linux-*
+          yum -y install make gcc-c++ flex bison git rpmdevtools wget zlib-devel
+          wget --no-verbose https://github.com/ccache/ccache/releases/download/v4.8.3/ccache-4.8.3-linux-x86_64.tar.xz
+          unxz ccache-4.8.3-linux-x86_64.tar.xz
+          tar xf ccache-4.8.3-linux-x86_64.tar
```
tautschnig, line 117: Use `tar xJf` instead of the separate `unxz` step?

.github/workflows/pull-request-checks.yaml
```diff
@@ -100,3 +100,45 @@ jobs:
...
+          tar xf ccache-4.8.3-linux-x86_64.tar 
+          cp ccache-4.8.3-linux-x86_64/ccache /usr/bin/
+          # yum install dnf-plugins-core
+          # yum config-manager --set-enabled powertools
+          # yum install glibc-static libstdc++-static
+      - name: cache for ccache
+        uses: actions/cache@v3
+        with:
+          path: /github/home/.ccache
+          key: ${{ runner.os }}-make-gcc-${{ github.ref }}-${{ github.sha }}-PR
```
tautschnig, line 126: I'd use `${{ runner.os }}-centos8-make-gcc-${{ github.ref }}-${{ github.sha }}-PR` for consistency.

.github/workflows/pull-request-checks.yaml
```diff
@@ -100,3 +100,45 @@ jobs:
...
+          restore-keys: ${{ runner.os }}-
```
tautschnig, line 127: Needs to be more specific, else cache with necessarily high cache-miss rate will be restored. Use `${{ runner.os }}-make-gcc-` (or `${{ runner.os }}-centos8-make-gcc` if my above proposal is adopted).

.github/workflows/pull-request-checks.yaml
```diff
@@ -100,3 +100,45 @@ jobs:
...
+          submodules: recursive
+      - name: Zero ccache stats and limit in size
+        run: ccache -z --max-size=500M
+      - name: Get minisat
+        run: make -C lib/cbmc/src minisat2-download
+      - name: Build with make
+        run: make CXX="ccache g++" -C src -j2
+      - name: Run the ebmc tests with SAT
+        run: |
+          rm regression/ebmc/neural-liveness/counter1.desc
```
tautschnig, line 139: Why does this one need to be removed (in CentOS only)?

.github/workflows/pull-request-checks.yaml
kroening: done

.github/workflows/pull-request-checks.yaml
kroening: ok

.github/workflows/pull-request-checks.yaml
kroening: ok

.github/workflows/pull-request-checks.yaml
kroening: Yet to be investigated; maybe the shell.

tautschnig: Ah, that test should likely use `echo -e`, or maybe use `printf` instead to have more portable support for backslash-n.

tautschnig (approved):

PR #168, opened 2023-11-13
Verilog: set values of parameter ports

kroening:


src/verilog/verilog_parameterize_module.cpp
```diff
@@ -31,19 +31,21 @@ verilog_typecheckt::get_parameter_declarators(const irept &module_source)
 {
   std::vector<verilog_parameter_declt::declaratort> declarators;
 
+  // We do the parameter ports first.
+  const irept &parameter_port_list = module_source.find(ID_parameter_port_list);
+
+  for(auto &decl : parameter_port_list.get_sub())
+    declarators.push_back(
+      static_cast<const verilog_parameter_declt::declaratort &>(decl));
```
tautschnig, line 39: This cast (and perhaps even more the C-style cast in line 199) suggest that we need a better API (or at least `to_X` functions) here.

tautschnig (approved):

src/verilog/verilog_parameterize_module.cpp
kroening: Yes, I am planning to use `verilog_module_exprt` for the module source as well, coming.

PR #169, opened 2023-11-14
introduce verilog_typecheck_exprt::hierarchical_identifier(base_name)

kroening:
This consolidates the code that forms the identifier for a symbol from a `base_name` in the context of a module, function/task within that module, and the named block within the module or function/task.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -27,6 +27,32 @@ Author: Daniel Kroening, kroening@kroening.com
 
 /*******************************************************************\
 
+Function: verilog_typecheck_exprt::hierarchical_identifier
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+irep_idt
+verilog_typecheck_exprt::hierarchical_identifier(irep_idt base_name) const
+{
+  const std::string named_block =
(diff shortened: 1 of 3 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -27,6 +27,32 @@ Author: Daniel Kroening, kroening@kroening.com
...
+  const irep_idt named_block =
+    named_blocks.empty() ? irep_idt() : named_blocks.back();
+
+  if(function_or_task_name.empty())
+    return id2string(module_identifier) + "." + id2string(named_block) +
+           id2string(base_name);
+  else
+    return id2string(function_or_task_name) + "." + id2string(named_block) +
+           id2string(base_name);
+}
```
tautschnig, line 52: Wouldn't it be cleaner to make more use of `std::string` here? E.g., use `const std::string &` as parameter type and make `named_block` a `const std::string`? I can see that the former will make callers use `id2string` instead, so perhaps that's not as good. But I'm pretty convinced that `named_block` should be typed differently.

tautschnig (approved):

src/verilog/verilog_typecheck_expr.cpp
kroening: I'll make `named_block` a string.

PR #178, opened 2023-11-14
ebmc: add --show-module-hierarchy

kroening:
This adds the command line option `--show-module-hierarchy` to ebmc, which outputs the hierarchy of modules, starting from the given top/main module.


src/trans-word-level/show_module_hierarchy.cpp
```diff
@@ -0,0 +1,66 @@
...
+      symbol.second.module == main_symbol.name)
+    {
+      instances.push_back(symbol.first);
+    }
+
+  // sort the instances alphabetically
+  std::sort(
+    instances.begin(),
+    instances.end(),
+    [](irep_idt a, irep_idt b) { return a.compare(b) < 0; });
```
tautschnig, line 41: Just ideas in case you want to pick it up one of them:
1. Use a `std::set<std::string>` to represent `instances` (that's what we do in a bunch of places in CBMC)
2. Use
````c++
auto comp = [](irep_idt a, irep_idt b) { return a.compare(b) < 0; };
std::set<irep_idt, decltype(comp)> instances(comp);
````

tautschnig (approved):

PR #179, opened 2023-11-15
introduce verilog_generate_blockt

kroening:
This introduces a class for generate block module items, to strengthen typing in the Verilog frontend.

### DIFF
--- a/src/verilog/verilog_expr.h
+++ b/src/verilog/verilog_expr.h
@@ -164,6 +164,59 @@ extern inline verilog_module_itemt &to_verilog_module_item(irept &irep)
   return static_cast<verilog_module_itemt &>(irep);
 }
 
+class verilog_generate_blockt : public verilog_module_itemt
+{
+public:
+  explicit verilog_generate_blockt(
+    verilog_module_exprt::module_itemst _module_items)
+    : verilog_module_itemt(ID_generate_block)
+  {
+    module_items() = std::move(_module_items);
+  }
+
+  verilog_generate_blockt(
+    irep_idt _identifier,
+    verilog_module_exprt::module_itemst _module_items)
+    : verilog_module_itemt(ID_generate_block)
+  {
+    set(ID_identifier, _identifier);
+    module_items() = std::move(_module_items);
+  }
+
+  const irep_idt &identifier() const
+  {
+    return get(ID_identifier);
+  }
+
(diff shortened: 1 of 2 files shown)


src/verilog/verilog_expr.h
```diff
@@ -164,6 +164,59 @@ extern inline verilog_module_itemt &to_verilog_module_item(irept &irep)
...
+  verilog_generate_blockt(
+    irep_idt _identifier,
+    verilog_module_exprt::module_itemst _module_items)
+    : verilog_module_itemt(ID_generate_block)
+  {
+    set(ID_identifier, _identifier);
+    module_items() = std::move(_module_items);
+  }
+
+  inline irep_idt identifier() const
```
tautschnig, line 186: Should this be a `const irep_idt &`? (If we ever were to turn of `USE_DSTRING`.)

tautschnig (approved):

src/verilog/verilog_expr.h
kroening: Sure. I am starting to wonder whether the noise is worth it.

src/verilog/verilog_expr.h
```diff
@@ -164,6 +164,59 @@ extern inline verilog_module_itemt &to_verilog_module_item(irept &irep)
...
+  verilog_module_exprt::module_itemst &module_items()
+  {
+    return (verilog_module_exprt::module_itemst &)operands();
+  }
+};
+
+extern inline const verilog_generate_blockt &
+to_verilog_generate_block(const exprt &expr)
+{
+  assert(expr.id() == ID_generate_block);
```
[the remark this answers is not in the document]
kroening: ok

src/verilog/verilog_expr.h
tautschnig: Quite probably the code base doesn't even compile without `USE_DSTRING` anymore. I'll add experimenting with this to my backlog. Perhaps a brief performance experiment will tell whether it is at all worth keeping this option.

kroening: The reason why `USE_DSTRING` exists in the first place was an implementation of symbolic execution that introduced a lot of fresh identifiers, thereby giving rise to a need to do free strings. I think that there are good alternatives to creating lots of fresh strings.

PR #180, opened 2023-11-15
introduce verilog_module_sourcet

kroening:
This documents an existing irep-based data structure used internally in the Verilog frontend.

### DIFF
--- a/src/verilog/verilog_expr.h
+++ b/src/verilog/verilog_expr.h
@@ -190,16 +190,16 @@ public:
     }
   };
 
-  using declarators = std::vector<declaratort>;
+  using declaratorst = std::vector<declaratort>;
 
-  const declarators &declarations() const
+  const declaratorst &declarations() const
   {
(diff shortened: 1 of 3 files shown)


src/verilog/verilog_expr.h
```diff
@@ -1132,4 +1132,59 @@ extern inline verilog_assumet &to_verilog_assume(exprt &expr)
   return static_cast<verilog_assumet &>(expr);
 }
 
+class verilog_module_sourcet : public irept
+{
+public:
+  irep_idt name() const
+  {
+    return get(ID_name);
+  }
```
tautschnig, line 1141: Should there be some constructor that sets field(s)? Perhaps at least the name?

tautschnig (approved):

src/verilog/verilog_expr.h
kroening: Will come -- I'll replace `verilog_modulet` by this class.

PR #181, opened 2023-11-15
Verilog: add a KNOWNBUG test for a parameter dependency

kroening:
Verilog parameters may depend on parameters declared later in the module.

### DIFF
--- a/regression/verilog/modules/parameters6.desc
+++ b/regression/verilog/modules/parameters6.desc
@@ -1,0 +1,8 @@
+KNOWNBUG
+parameters6.v
+
+^EXIT=0$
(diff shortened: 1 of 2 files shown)


regression/verilog/modules/parameters6.desc
```diff
@@ -0,0 +1,6 @@
+KNOWNBUG
+parameters6.v
+
+^EXIT=0$
+^SIGNAL=0$
+--
```
tautschnig, line 6: I'd appreciate if "KNOWNBUG" tests include (after a further `--` line) a comment that says what gap in our implementation this test refers to.

PR #186, opened 2023-11-20
Verilog: use verilog_module_sourcet when checking the ports

kroening:
This strengthens the typing in the Verilog front-end.

### DIFF
--- a/src/verilog/verilog_interfaces.cpp
+++ b/src/verilog/verilog_interfaces.cpp
@@ -34,13 +34,13 @@ void verilog_typecheckt::module_interface(
   for(auto &module_item : module_source.module_items())
     interface_module_item(module_item);
 
-  // now check port names
-  interface_ports(module_symbol.type.add(ID_ports).get_sub());
+  // Check the typing of the ports
+  check_module_ports(module_source);
 }
 
 /*******************************************************************\
 
-Function: verilog_typecheckt::interface_ports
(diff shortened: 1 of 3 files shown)


src/verilog/verilog_interfaces.cpp
```diff
@@ -50,17 +50,18 @@ Function: verilog_typecheckt::interface_ports
...
+void verilog_typecheckt::check_module_ports(
+  const verilog_module_sourcet &module_source)
 {
-  const irept &module_source=module_symbol.type.find(ID_module_source);
-  const irept &module_ports=module_source.find(ID_ports);
+  const auto &module_ports = module_source.ports();
 
-  ports.resize(module_ports.get_sub().size());
+  auto &ports = module_symbol.type.add(ID_ports).get_sub();
+  ports.resize(module_ports.size());
```
tautschnig, line 59: How about getting rid of the `resize` here and change it as follows:
1. Use `reserve` here.
2. Get rid of `nr`.
3. Use `push_back` instead of access via `[nr]`.
4. Use `port_names[name]=ports.size() - 1;`

That would avoid any risk of having uninitialised ports in here.

tautschnig (approved):

src/verilog/verilog_interfaces.cpp
kroening: ok, will make separate PR

PR #187, opened 2023-11-20
Verilog: introduce elaborate_generate_decl

kroening:
This extracts a method to process declarations during the elaboration of generate constructs.

### DIFF
--- a/src/verilog/verilog_generate.cpp
+++ b/src/verilog/verilog_generate.cpp
@@ -61,6 +61,44 @@ void verilog_typecheckt::elaborate_generate_block(
 
 /*******************************************************************\
 
+Function: verilog_typecheckt::elaborate_generate_decl
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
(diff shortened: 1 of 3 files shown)


src/verilog/verilog_generate.cpp
```diff
@@ -61,6 +61,43 @@ void verilog_typecheckt::elaborate_generate_block(
...
+void verilog_typecheckt::elaborate_generate_decl(
+  const verilog_declt &decl,
+  module_itemst &dest)
+{
+  auto decl_class = decl.get_class();
+
+  if(decl_class == ID_genvar)
+  {
+    for(auto &declarator : decl.declarators())
+      genvars[declarator.identifier()] = -1;
```
tautschnig, line 83: I'm aware that this just code that has been moved around, but still: would it be possible to add a comment explaining the "-1" magic number?

tautschnig (approved):

src/verilog/verilog_generate.cpp
kroening: done

PR #188, opened 2023-11-20
Verilog: add symbols for variables in generate constructs

kroening:
This fixes the Verilog front-end to add symbols for variables that are declared in generate constructs.


tautschnig (approved):
Not merging (yet) as #187 should be merged first.

PR #196, opened 2023-11-24
Verilog: typedefs inside modules

kroening:
Following IEEE 1800 2017, this allows typedefs as `block_item` or `package_or_generate_item`.


tautschnig (approved):
Needs a rebase now that rules have been shuffled around in the other PRs.

PR #198, opened 2023-11-25
Verilog: use data_declaration for reg, event, integer, real declarations

kroening:


regression/verilog/data-types/non-integer-types.sv
```diff
@@ -1,8 +1,8 @@
 module main();
 
    // IEEE 1800-2017 A.2.2.1
-   reg shortreal some_shortreal;
-   reg real some_real;
-   reg realtime some_realtime;
+   shortreal some_shortreal;
```
tautschnig, line 4: Why was it useful/necessary to remove `reg` here?

tautschnig (approved):

regression/verilog/data-types/non-integer-types.sv
kroening: Yes, perhaps surprisingly so, the Verilog standard considers `reg` a type.

PR #200, opened 2023-11-25
Verilog enum

kroening:


src/verilog/verilog_elaborate_constants.cpp
```diff
@@ -1,144 +1,271 @@
...
+  auto full_identifier =
+    id2string(module_identifier) + "." + id2string(base_name);
 
-  Inputs:
+  // If there's no type, parameters take the type of the
+  // value. We signal this using the special type "derive_from_value".
 
- Outputs:
+  auto symbol_type = type_with_subtypet(
+    ID_to_be_elaborated, type.is_nil() ? typet("derive_from_value") : type);
```
tautschnig, line 30: Introduce a new `ID_derive_from_value`?

src/verilog/verilog_elaborate_constants.cpp
```diff
@@ -1,144 +1,271 @@
...
+    const auto &enum_type = to_verilog_enum_type(type);
+
+    if(enum_type.has_base_type())
+      collect_symbols(enum_type.base_type());
+
+    /*
+    // In contrast to C, there is no tag in the syntax, but our parser
+    // generates one. We use that as type for the enum names.
+    const verilog_enum_tag_typet enum_tag_type(enum_type.identifier());
+    */
```
tautschnig, line 59: What to make of this commented-out block?

src/verilog/verilog_elaborate_constants.cpp
```diff
@@ -1,144 +1,271 @@
...
 
-  std::vector<irep_idt> to_be_elaborated;
+    // The default base type is 'int'.
+    auto base_type =
+      enum_type.has_base_type() ? enum_type.base_type() : signedbv_typet(32);
+
+    // Add the enum names to the symbol table for subsequent elaboration.
+    // Values are given, or the previous plus one, starting with value '0'.
+    exprt initializer = from_integer(0, integer_typet());
```
tautschnig, line 67: Why not use `base_type` instead of `integer_typet`?

src/verilog/verilog_typecheck_type.cpp
```diff
@@ -74,6 +74,25 @@ typet verilog_typecheck_exprt::convert_type(const typet &src)
...
   }
+  else if(src.id() == ID_verilog_enum)
+  {
+    /*
+    // Replace by enum tag type.
+    typet result =
+      verilog_enum_tag_typet(to_verilog_enum_type(src).identifier());
+    result.add_source_location() = std::move(source_location);
+    return result;
+    */
```
tautschnig, line 85: What to make of the comment here?

tautschnig (approved):

src/verilog/verilog_elaborate_constants.cpp
kroening: That one is subtle. I will add a comment. The base type might not yet have been elaborated at this time yet. I might move this code to happen after that elaboration, but that requires the ability to add further symbols during elaboration.

src/verilog/verilog_elaborate_constants.cpp
kroening: It's one of the KNOWNBUG tests -- enums in Verilog are meant to be strictly typed, and I am wondering whether to achieve that with a tag. But there's likely no need, given that the enum names must be unique.

src/verilog/verilog_elaborate_constants.cpp
kroening: done

PR #204, opened 2023-11-27
Verilog: transition to structured error exception

kroening:
This uses a copy of `typecheckt::errort`, which can be removed when upgrading the cbmc dependency.  The exceptions thrown in `verilog_typecheck_exprt` are changed to the new exception type.


tautschnig (approved):
What about upgrading the CBMC dependency right away? We use git submodules, so don't depend on releases to happen?

kroening:
Will do separate PR.

PR #205, opened 2023-11-27
Verilog: typedef

kroening:
This enables distinguishing type identifiers from other identifiers while parsing, which enables System Verilog `typedef`.


src/verilog/verilog_typecheck_type.cpp
```diff
@@ -59,6 +59,21 @@ typet verilog_typecheck_exprt::convert_type(const typet &src)
...
+    // Look it up!
+    const symbolt *symbol_ptr;
+
+    auto identifier = src.get(ID_identifier);
+
+    if(ns.lookup(identifier, symbol_ptr))
+      throw errort().with_location(source_location)
+        << "type symbol " << identifier << " not found";
+
+    DATA_INVARIANT(symbol_ptr->is_type, "typedef symbols must be types");
```
tautschnig, line 73: Is this really something that we don't expect user code to trigger? Could it not be triggered by using a signal's name in a place where a type is required?

tautschnig (approved):

src/verilog/verilog_typecheck_type.cpp
kroening: That would cause a syntax error.

PR #208, opened 2023-11-29
Verilog: synth_expr now returns the result

kroening:
Return the result of expression synthesis instead of modifying in place.

### DIFF
--- a/src/verilog/verilog_synthesis.cpp
+++ b/src/verilog/verilog_synthesis.cpp
@@ -38,9 +38,7 @@ Function: verilog_synthesist::synth_expr
 
 \*******************************************************************/
 
-void verilog_synthesist::synth_expr(
-  exprt &expr,
-  symbol_statet symbol_state)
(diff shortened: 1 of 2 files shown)


src/verilog/verilog_synthesis.cpp
```diff
@@ -53,60 +51,72 @@ void verilog_synthesist::synth_expr(
...
-        break;
+        return final_value(symbol);
 
       case symbol_statet::NONE:
         error().source_location=expr.source_location();
         error() << "symbols not allowed here" << eom;
         throw 0;
       }
+
+      UNREACHABLE;
```
tautschnig, line 83: Should be unnecessary as we compile with warnings about missing cases.

tautschnig (approved):

src/verilog/verilog_synthesis.cpp
kroening: Neither clang nor gcc worked that out.

kroening: Correction: clang does work that out, gcc doesn't.

PR #226, opened 2023-12-02
ebmc: fix encoding of 'eventually'

kroening:
This strengthens the constraints for `eventually p` to prevent wrong counterexamples where `p` is satisfied somewhere on the lasso.


tautschnig (approved):
Requires a rebase.

PR #227, opened 2023-12-02
ebmc: add waveform output

kroening:
This adds a "horizontal" output format for traces, where the timeframe is on the x axis and the identifier of the state variable are on the y axis.

### DIFF
--- a/regression/ebmc/traces/waveform1.desc
+++ b/regression/ebmc/traces/waveform1.desc
@@ -1,0 +1,12 @@
+CORE broken-smt-backend
+waveform1.smv
+--bound 20 --waveform
+^EXIT=10$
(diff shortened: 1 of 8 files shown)


tautschnig (approved):
1. Could this please be mentioned in `--help` output?
2. Is this format understood by other tools (gtkwave?) or otherwise standardized? If so, it would be great to mention this somewhere.

kroening:
1. Done
2. gtkwave reads VCD!

tautschnig:
[...]
>     2. gtkwave reads VCD!

That's what I thought so I was wondering whether it perhaps also supported such an input, or whether some other tool did so? If we are making up our own format that's surely ok, but I think it might be worth explicitly noting this.

PR #228, opened 2023-12-02
ebmc: add a test for randomly generating a trace with 1000 steps

kroening:
This adds a test for randomly generating a trace with 1000 steps.


kroening:
It is arbitrary; I got worried when I saw that all tests exercised two states at most.

PR #229, opened 2023-12-02
ebmc: show-formula now includes lasso constraints

kroening:
Add the lasso constraints when showing the generated formula.


kroening:
> I think this one could do with a test.

done

PR #231, opened 2023-12-02
Remove dependency zlib

mgudemann:
the existing dependency on `zlib` seems to be unnecessary.
`zlib.h` is only included in the `ic3` Minisat and applying similar patches to what is used in CBMC removes this dependency.
In the long run, `ic3` Minisat should maybe be replaced with CBMC Minisat


mgudemann:
Done. I had just commented-out the lines as this is how it's done in the minisat patchset for CBMC

PR #234, opened 2023-12-03
Sort definitions of SMV models to reduce recursion depth in typecheck

### LINKED ISSUE #17: segfault on typecheck
For some of the SMV models converted from Aiger format, EBMC fails with sefgault which typechecking.
The example is from http://www.cprover.org/ebmc/download/AIG_SMV_08.tar.gz

[aig.zip](https://github.com/diffblue/hw-cbmc/files/1117215/aig.zip)

mgudemann:
This fixes #17 (superseeding #26 as EBMC code has changed)

### DIFF
--- a/src/smvlang/parser.y
+++ b/src/smvlang/parser.y
@@ -11,7 +11,7 @@
 
 #include "smv_y.tab.h"
 
-#define YYMAXDEPTH 200000
+#define YYMAXDEPTH 2000000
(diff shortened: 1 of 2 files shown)


kroening:
How about using a `std::stack`?

mgudemann:
For the topological sort? Sure, but this is implemented in `cbmc/src/util/graph.h` so it would require a change to CBMC.

kroening:
#1040

PR #241, opened 2023-12-04
Verilog: add a KNOWNBUG test for a for loop over a register

kroening:


regression/verilog/for/for_with_reg.desc
```diff
@@ -0,0 +1,7 @@
+CORE
```
tautschnig, line 1: Should this be `KNOWNBUG`?

kroening: Well, it shouldn't be!

tautschnig (approved):

PR #253, opened 2023-12-05
ebmc: add --random-waveform and --random-trace

kroening:
This adds two command-line options `--random-waveform` and `--random-trace`, which print a single random trace to the console.


tautschnig (approved):
clang-format has some suggestions.

PR #263, opened 2023-12-08
ebmc: transition_systemt does not need optional transt

kroening:
The `trans_expr` in `transition_systemt` is always populated, and does not need to be optional.

### DIFF
--- a/src/ebmc/transition_system.h
+++ b/src/ebmc/transition_system.h
@@ -10,6 +10,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #define CPROVER_EBMC_TRANSITION_SYSTEM_H
 
 #include <util/mathematical_expr.h>
+#include <util/std_expr.h>
 #include <util/symbol_table.h>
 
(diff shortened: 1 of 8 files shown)


src/ebmc/transition_system.h
```diff
@@ -18,9 +19,19 @@ class message_handlert;
...
 public:
+  transition_systemt()
+    : trans_expr{
+        ID_trans,
+        conjunction({}),
+        conjunction({}),
+        conjunction({}),
+        typet()}
+  {
+  }
```
tautschnig, line 30: 1.  Is there some place where this should now be used?
2. I am on the fence whether the empty conjunction is helpful here or not: it makes clear that these operands should eventually hold conjunctions, but does hide that this is just a fancy way of producing `true_exprt`.

tautschnig (approved):

src/ebmc/transition_system.h
kroening: Replaced by `true_exprt()`

PR #272, opened 2023-12-10
ebmc: avoid parentheses in SVA output for unary operators

kroening:
This removes parentheses when printing unary SVA operators that have operands that are unary.


tautschnig:
Needs a rebase.

tautschnig (approved):

PR #273, opened 2023-12-10
ebmc: re-enable IC3

kroening:


regression/ebmc/ic3/pdtvispeterson.desc
```diff
@@ -1,6 +1,8 @@
 CORE
 pdtvispeterson.sv
 --ic3
+^EXIT=2$
+^SIGNAL=0$
 ^property HOLDS
 ^inductive invariant verification is ok
```
tautschnig, line 7: What do these exit codes (here: 2, in other tests: 1) mean, and why is the result output seemingly so different?

kroening: The result reporting is very different, both in terms of exit code and console output, and it's a todo item to align it. The changes to the test merely document the existing behaviour.

tautschnig (approved):

PR #275, opened 2023-12-11
Update COMPILING

hayat10x:
for clearing completed the sentence


tautschnig:
Thank you for pointing out the lack of clarity. With #278, however, we have even simpler instructions. (There really is no need to separately compile CBMC).

PR #277, opened 2023-12-11
ebmc: error k-induction when given liveness property

kroening:
Return an error when k-induction is given a liveness property.

### DIFF
--- a/src/ebmc/k_induction.cpp
+++ b/src/ebmc/k_induction.cpp
@@ -14,6 +14,7 @@ Author: Daniel Kroening, daniel.kroening@inf.ethz.ch
 #include <trans-word-level/unwind.h>
 
 #include "ebmc_base.h"
+#include "ebmc_error.h"
 #include "ebmc_solver_factory.h"
 #include "liveness_to_safety.h"
 #include "report_results.h"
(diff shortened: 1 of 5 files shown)


tautschnig:
Needs a rebase.

kroening:
done

src/ebmc/k_induction.cpp
```diff
@@ -90,14 +91,20 @@ int k_inductiont::operator()()
...
   if(cmdline.isset("liveness-to-safety"))
     liveness_to_safety(transition_system, properties);
 
-  if(properties.properties.empty())
+  // Check whether the properties are suitable for k-induction.
+  if(properties.requires_lasso_constraints())
   {
-    message.error() << "no properties" << messaget::eom;
-    return 1;
+    throw ebmc_errort() << "k-induction does not support liveness properties";
```
tautschnig, line 107: I think it would be useful if this error included a source location (e.g., the source location of the first liveness property that we find. This, of course, requires adjusting the `requires_lasso_constraints` interface so might be a change for a future PR, but would still make it easier for a user.

tautschnig (approved):

src/ebmc/k_induction.cpp
kroening: Now with source location

PR #292, opened 2023-12-13
Verilog: this

kroening:
This adds support for the System Verilog `this` keyword.

### DIFF
--- a/regression/verilog/expressions/this1.desc
+++ b/regression/verilog/expressions/this1.desc
@@ -1,0 +1,7 @@
+CORE
+this1.sv
+
+^file .* line 3: 'this' outside of method$
+^EXIT=2$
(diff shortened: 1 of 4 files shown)


regression/verilog/expressions/this1.desc
```diff
@@ -0,0 +1,7 @@
+CORE
+this1.sv
+
+^file .* line 3: 'this' outside of method$
```
tautschnig, line 4: Shouldn't we also have a test where it is actually permitted?

tautschnig (approved):

regression/verilog/expressions/this1.desc
kroening: This would need an implementation of classes and methods.

PR #294, opened 2023-12-13
Verilog: System Verilog typecasts

kroening:
This adds support for System Verilog typecast expressions.

### DIFF
--- a/regression/verilog/casts/static_cast1.desc
+++ b/regression/verilog/casts/static_cast1.desc
@@ -1,0 +1,8 @@
+CORE
+static_cast1.sv
+--module main --bound 0
+^EXIT=0$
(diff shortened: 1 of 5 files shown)


regression/verilog/casts/static_cast1.desc
```diff
@@ -0,0 +1,8 @@
+KNOWNBUG
```
tautschnig, line 1: Is this still correct? If so, could you please add a comment at the end

tautschnig (approved):

regression/verilog/casts/static_cast1.desc
kroening: Ah, no, actually works, will fix

PR #307, opened 2023-12-18
Verilog: enable simplification of concatenations with signed operands

kroening:
This adds a cast to signed operands of concatenation expressions, to enable simplification.


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -276,6 +276,19 @@ exprt verilog_typecheck_exprt::convert_expr_rec(exprt expr)
...
+    // We cast all the signed operands to unsigned.
+    for(auto &op : expr.operands())
+    {
+      if(op.type().id() == ID_signedbv || op.type().id() == ID_verilog_signedbv)
+      {
+        auto width = get_width(op);
+        auto op_type = op.type().id() == ID_verilog_signedbv
+                         ? static_cast<typet>(verilog_unsignedbv_typet(width))
+                         : static_cast<typet>(unsignedbv_typet(width));
+        op = typecast_exprt(op, op_type);
```
tautschnig, line 288: How about:
````c++
if(op.type().id() == ID_verilog_signedbv)
  op = typecast_exprt(op, verilog_unsignedbv_typet(width));
else
  op = typecast_exprt(op, unsignedbv_typet(width));
````
to avoid the `static_cast` plus the constructed `typet` will get moved instead of being copied.

tautschnig (approved):

PR #308, opened 2023-12-19
bump cbmc dependency

kroening:

PR #315, opened 2023-12-23
add constraints for assignments to function-local variables

kroening:
Assignments to function-local variables are now treated as cycle-local, and hence neither require invariants nor next-state constraints.


src/verilog/verilog_synthesis.cpp
```diff
@@ -112,6 +111,32 @@ exprt verilog_synthesist::synth_expr(exprt expr, symbol_statet symbol_state)
...
+\*******************************************************************/
+
+void verilog_synthesist::function_locality(const symbolt &function_symbol)
+{
+  // Find all local symbols of the function, mark their
+  // assignments as local.
+  auto prefix = id2string(function_symbol.name) + '.';
+  for(auto &symbol : symbol_table.symbols)
+  {
+    if(irep_idt(symbol.first).starts_with(prefix))
```
tautschnig, line 131: Should we perhaps get the latest version of the CBMC dependency merged first so that this `irep_idt(...)` wrapper can be avoided?

src/verilog/verilog_synthesis.cpp
```diff
@@ -362,22 +390,28 @@ void verilog_synthesist::assignment_rec(
...
+    {
+      if(assignment.type == event_guardt::NONE)
+        assignment.type = new_type;
+      else if(assignment.type != new_type)
+      {
+        error().source_location = lhs.source_location();
+        error() << "conflicting assignment types for `" << symbol.base_name
+                << "' (new: " << as_string(new_type)
+                << ", old: " << as_string(assignment.type) << ")" << eom;
+        throw 0;
```
tautschnig, line 406: Use `errort`?

tautschnig (approved):

src/verilog/verilog_synthesis.cpp
kroening: #308

PR #316, opened 2023-12-24
typed parameter ports

kroening:


regression/verilog/modules/parameter_ports2.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
```
tautschnig, line 1: Is this still meant to be a `KNOWNBUG`?

tautschnig (approved):

regression/verilog/modules/parameter_ports2.desc
kroening: Yes, the type still needs to be casted.

PR #317, opened 2023-12-24
Verilog: --preprocess now honors -I

kroening:


src/ebmc/transition_system.cpp
```diff
@@ -92,6 +92,10 @@ int preprocess(const cmdlinet &cmdline, message_handlert &message_handler)
     return 1;
   }
 
+  // do -I
+  if(cmdline.isset('I'))
+    config.verilog.include_paths = cmdline.get_values('I');
+
```
tautschnig, line 98: Given that it seemed surprising that no one noticed this didn't work: could you please add a test?

tautschnig (approved):

PR #326, opened 2023-12-26
Verilog: verilog_ift interface

kroening:
The interface of `verilog_ift` is changed to match the interface of `code_ifthenelset`.


tautschnig:
Needs a rebase.

src/verilog/verilog_expr.h
```diff
@@ -784,51 +784,73 @@ inline verilog_caset &to_verilog_case(exprt &expr)
...
-  exprt &condition()
+  verilog_ift(
+    exprt __cond,
+    verilog_statementt __then_case,
+    verilog_statementt __else_case)
+    : verilog_statementt(ID_if)
+  {
+    add_to_operands(
+      std::move(__cond), std::move(__then_case), std::move(__else_case));
+  }
```
tautschnig, line 801: Just a thought: would it be simpler to always use 3 operands and, when necessary, make that third operand `nil`?

tautschnig (approved):

src/verilog/verilog_expr.h
kroening: Possibly; but a goal was to match `code_ifthenelset`, which would then also have to change.

PR #331, opened 2024-01-09
function calls can be constants

kroening:
Function calls with constant arguments can be elaboration time constants. This moves the creation of the function symbols into the constant elaboration phase to allow this.


kroening:
Superseded by #361

PR #339, opened 2024-01-13
introduce classes for SVA expressions

kroening:


src/ebmc/ebmc_properties.cpp
```diff
@@ -143,7 +144,7 @@ ebmc_propertiest ebmc_propertiest::from_command_line(
 
     if(expr.id() != ID_sva_always)
     {
-      unary_predicate_exprt tmp(ID_sva_always, expr);
+      sva_always_exprt tmp(expr);
       expr.swap(tmp);
```
tautschnig, line 148: Wouldn't `expr = sva_always_exprt{expr};` also do the trick, i.e., no need for `swap`?

tautschnig (approved):

PR #343, opened 2024-01-15
add KNOWNBUG tests

kroening:


regression/verilog/elaboration/type_operator.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+type_operator.sv
+--bound 0
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
```
tautschnig, line 7: Would it be possible to add (here and in the other tests being added) a note on what is going wrong, or perhaps even an idea how the problem will be fixed?

tautschnig (approved):

PR #347, opened 2024-01-15
nuterm: add 'live' signal

kroening:


src/ebmc/random_traces.cpp
```diff
@@ -320,6 +320,31 @@ void random_traces(
...
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+void random_traces(
```
tautschnig, line 333: Ah, we never had that implementation :-)

PR #348, opened 2024-01-15
Verilog: replace equality by declarator class

kroening:
This replaces the use of an equality expression for declarators with initializers by the declarator class.

### DIFF
--- a/src/verilog/verilog_expr.h
+++ b/src/verilog/verilog_expr.h
@@ -308,6 +308,11 @@ public:
       return get(ID_identifier);
     }
 
+    void identifier(irep_idt _identifier)
+    {
+      set(ID_identifier, _identifier);
+    }
+
     const irep_idt &base_name() const
     {
       return get(ID_identifier);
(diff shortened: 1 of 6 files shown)


src/verilog/verilog_expr.h
```diff
@@ -322,6 +327,13 @@ class verilog_parameter_declt : public verilog_module_itemt
     {
       return static_cast<exprt &>(add(ID_value));
     }
+
+    // helper to generate a symbol expression
+    symbol_exprt symbol_expr() const
+    {
+      return symbol_exprt(identifier(), type())
```
tautschnig, line 334: Might use `return symbol_exprt::typeless(identifier())`

tautschnig (approved):

src/verilog/verilog_expr.h
kroening: Ah, note it's `type()`, not `typet()`. It's the type of the declarator, not the empty type.

PR #349, opened 2024-01-16
Verilog: use declaratort instead of symbol_exprt

kroening:
This replaces the use of `symbol_exprt` as a declarator in all declarations by the dedicated type `declaratort`.

### DIFF
--- a/src/verilog/verilog_interfaces.cpp
+++ b/src/verilog/verilog_interfaces.cpp
@@ -310,28 +310,17 @@ void verilog_typecheckt::interface_function_or_task_decl(const verilog_declt &de
 
   for(auto &declarator : decl.declarators())
   {
-    if(declarator.id() == ID_symbol)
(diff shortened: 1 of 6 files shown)


src/verilog/verilog_interfaces.cpp
```diff
@@ -310,29 +310,18 @@ void verilog_typecheckt::interface_function_or_task_decl(const verilog_declt &de
...
-    {
-      throw errort().with_location(declarator.source_location())
-        << "unexpected declaration: " << declarator.id();
-    }
+    DATA_INVARIANT(declarator.id() == ID_declarator, "must have declarator");
+
+    symbol.base_name = declarator.base_name();
 
-    if(symbol.base_name=="")
+    if(symbol.base_name == "")
```
tautschnig, line 317: Nit pick: I'd use `.empty()`just in case there is a more efficient way in `std::string`to check that than the strcmp that the `== ""`will require.

tautschnig (approved):

PR #350, opened 2024-01-19
Verilog: genvar symbols now created during elaboration

kroening:
This moves the creation of genvar symbols from verilog_interfaces to the constant elaboration phase.


src/verilog/verilog_elaborate_constants.cpp
```diff
@@ -251,6 +251,33 @@ void verilog_typecheckt::collect_symbols(const verilog_declt &decl)
       }
     }
   }
+  else if(decl_class == ID_verilog_genvar)
+  {
+    symbolt symbol;
+    typet type;
```
tautschnig, line 257: Where is this used?

src/verilog/verilog_elaborate_constants.cpp
```diff
@@ -251,6 +251,33 @@ void verilog_typecheckt::collect_symbols(const verilog_declt &decl)
       }
     }
   }
+  else if(decl_class == ID_verilog_genvar)
+  {
+    symbolt symbol;
```
tautschnig, line 256: Perhaps use `symbolt symbol{irep_idt{}, verilog_genvar_typet{}, mode};`

tautschnig (approved):

src/verilog/verilog_elaborate_constants.cpp
kroening: removed

PR #359, opened 2024-01-30
Verilog: elaborate wire symbols

kroening:
This moves the creation of wire symbols from verilog_interfaces to the elaboration phase.


src/verilog/verilog_elaborate.cpp
```diff
@@ -369,6 +369,71 @@ void verilog_typecheckt::collect_symbols(const verilog_declt &decl)
...
+
+        // change of type?
+        if(symbol.type != osymbol.type)
+        {
+          if(get_width(symbol.type) > get_width(osymbol.type))
+            osymbol.type = symbol.type;
+        }
+      }
+
+      symbols_added.push_back(symbol.name);
```
tautschnig, line 435: Shouldn't this only be done when `result == nullptr`?

tautschnig (approved):
See also comment on the other PR.

PR #360, opened 2024-01-30
Verilog: elaborate var/reg

kroening:
This moves the creation of symbols for var/reg identifiers from verilog_interfaces to the elaboration phase.

### DIFF
--- a/src/verilog/verilog_elaborate.cpp
+++ b/src/verilog/verilog_elaborate.cpp
@@ -278,6 +278,24 @@ void verilog_typecheckt::collect_symbols(const verilog_declt &decl)
       add_symbol(symbol);
     }
   }
+  else if(
+    decl_class == ID_wire || decl_class == ID_supply0 ||
+    decl_class == ID_supply1 || decl_class == ID_triand ||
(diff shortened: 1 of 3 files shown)


src/verilog/verilog_elaborate.cpp
```diff
@@ -288,6 +290,78 @@ void verilog_typecheckt::collect_symbols(const verilog_declt &decl)
...
+        // a register can't be an input as well
+        if(osymbol.is_input && osymbol.is_state_var)
+        {
+          throw errort().with_location(decl.source_location())
+            << "symbol `" << symbol.base_name
+            << "' is declared both as input and as register";
+        }
+      }
+
+      symbols_added.push_back(symbol.name);
```
tautschnig, line 363: Is this correct when `result != nullptr`?

tautschnig (approved):

src/verilog/verilog_elaborate.cpp
kroening: Yes; having said that, I should add some test cases that cover a good subset of both the allowed and disallowed cases of redeclarations.

PR #371, opened 2024-02-10
Verilog: strong typing for enums

kroening:
The SystemVerilog 1800-2017 standard requires that enums are strongly typed, i.e., that automatic conversion to an enum type is prohibited.


src/verilog/verilog_parser.h
```diff
@@ -118,7 +117,9 @@ class verilog_parsert:public parsert
     return scope_ptr == nullptr ? false : scope_ptr->is_type;
   }
 
-  unsigned dummy_id;
+  // These are used for anonymous gate instances
+  // and to create a unique identifier for enum types.
+  std::size_t dummy_id = 0;
```
tautschnig, line 122: How about `next_unique_id`? (I'd also find it important to know whether the particular value is ok to use, or has been used already. Using "next" as part of the name would convey this information.)

src/verilog/verilog_parser.h
```diff
@@ -118,7 +117,9 @@ class verilog_parsert:public parsert
...
 
   std::string get_dummy_id()
```
tautschnig, line 124: How about also changing the name of this method, in keeping with my comment about the `dummy_id` variable?

tautschnig (approved):

src/verilog/verilog_parser.h
kroening: I'd hesitate to stick the word "unique" in there. The counter won't be unique (it gets reset to zero with each parser invocation). It only becomes unique once you add a suitable prefix. Renaming to `next_id_counter`.

PR #378, opened 2024-02-21
Verilog: enum constants may depend on elaboration-time constants

kroening:
This postpones the conversion of enum base types and enum initializers to enable them to depend on other elaboration-time constants.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -105,7 +105,8 @@ void verilog_typecheck_exprt::propagate_type(
       expr.type().get(ID_C_identifier) != type.get(ID_C_identifier))
     {
       throw errort().with_location(expr.source_location())
-        << "assignment to enum requires enum of the same type";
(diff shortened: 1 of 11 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1344,7 +1349,16 @@ exprt verilog_typecheck_exprt::elaborate_constant_expression(exprt expr)
 
     // We fall back to the simplifier to approximate
     // the standard's definition of 'constant expression'.
-    return simplify_expr(expr, ns);
+    auto simplified_expr = simplify_expr(expr, ns);
```
tautschnig, line 1352: Because we've been bitten by this in CBMC (cf. https://github.com/diffblue/cbmc/pull/7856): any use of the simplifier in the front-end needs to make sure that the resulting expression still has a source location.

tautschnig (approved):

src/verilog/verilog_typecheck_expr.cpp
kroening: added

PR #387, opened 2024-02-29
ebmc: HWMCC 2017 benchmarks in CI

kroening:
This adds a CI job for running the HWMCC 2017 safety benchmarks in CI.


benchmarking/hwmcc17.sh
```diff
@@ -0,0 +1,88 @@
+#!/bin/sh
+
+# ca. 30 minutes
```
tautschnig, line 3: What about using GNU parallel to run 4 jobs concurrently? Would that bring down execution time to almost a fourth? If so, I'd feel ok to have this in CI. Else, perhaps use a cron job to run it on a daily basis?

tautschnig (approved):

benchmarking/hwmcc17.sh
kroening: It should parallelise well; I'll try.

PR #389, opened 2024-03-01
ebmc: reporting when unsupported property is given

kroening:
This changes the behavior when an unsupported property is given to the BMC engine.  Instead of exiting, this is now reported in the property list. Supported properties are still checked.

### DIFF
--- a/src/ebmc/bmc.cpp
+++ b/src/ebmc/bmc.cpp
@@ -46,6 +46,13 @@ void bmc(
     if(property.is_disabled() || property.is_failure())
       continue;
 
+    // Is it supported by the BMC engine?
+    if(!bmc_supports_property(property.expr))
+    {
+      property.failure("property not supported by BMC engine");
+      continue;
+    }
+
     ::property(
       property.expr,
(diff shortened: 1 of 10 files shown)


src/ebmc/bmc.cpp
```diff
@@ -46,6 +46,13 @@ void bmc(
     if(property.is_disabled() || property.is_failure())
       continue;
 
+    // Is it supported by the BMC engine?
+    if(!bmc_supports_property(property.expr))
+    {
+      property.failure("property not supported by BMC engine");
+      continue;
+    }
+
```
tautschnig, line 55: What will be the exit code when all other properties are successfully proved? The regression tests all seem to have refuted properties.

tautschnig (approved):

src/ebmc/bmc.cpp
kroening: It should be a failure exit code, but I'll add the test for that.

kroening: Done

PR #394, opened 2024-03-03
Verilog: split assert/assume module items and statements

kroening:
SystemVerilog assertions and assumptions can be module items (called concurrent assertions/assumptions) or statements.  While the syntax for the two variants is similar, there are differences.  This commit splits the ireps for the two variants.

### DIFF
--- a/src/verilog/verilog_typecheck.cpp
+++ b/src/verilog/verilog_typecheck.cpp
@@ -1050,15 +1050,10 @@ Function: verilog_typecheckt::convert_assert
 
 \*******************************************************************/
 
-void verilog_typecheckt::convert_assert(exprt &statement)
+void verilog_typecheckt::convert_assert(
+  verilog_assert_module_itemt &module_item)
 {
-  if(statement.operands().size()!=2)
-  {
-    throw errort().with_location(statement.source_location())
-      << "assert statement expected to have two operands";
-  }
-
-  exprt &cond = to_binary_expr(statement).op0();
+  exprt &cond = module_item.condition();
 
   convert_expr(cond);
   make_boolean(cond);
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_typecheck.cpp
```diff
@@ -1123,15 +1118,95 @@ Function: verilog_typecheckt::convert_assume
...
+  else
+    property = sva_always_exprt(cond);
+
+  assertion_counter++;
+
+  const irep_idt &identifier = statement.identifier();
+
+  irep_idt base_name;
+
+  if(identifier == irep_idt())
```
tautschnig, line 1163: Use `.empty()`?!

src/verilog/verilog_typecheck.cpp
```diff
@@ -1072,12 +1067,12 @@ void verilog_typecheckt::convert_assert(exprt &statement)
...
-  const irep_idt &identifier=statement.get(ID_identifier);
-  
+
+  const irep_idt &identifier = module_item.identifier();
+
   irep_idt base_name;
-  
-  if(identifier=="")
+
+  if(identifier == irep_idt())
```
tautschnig, line 1075: Why not use `.empty()`?

src/verilog/verilog_typecheck.cpp
```diff
@@ -1123,15 +1118,95 @@ Function: verilog_typecheckt::convert_assume
...
   {
     throw errort().with_location(statement.source_location())
-      << "assume statement expected to have two operands";
+      << "property identifier `" << base_name << "' already used";
   }
 
-  exprt &cond = to_binary_expr(statement).op0();
+  statement.identifier(full_identifier);
+
+  symbolt symbol;
```
tautschnig, line 1179: Use `symbolt symbol{base_name, bool_typet{}, mode};`

tautschnig (approved):

PR #410, opened 2024-03-11
ebmc: add --system-verilog command-line option

kroening:
This adds the command-line option `--system-verilog` to ebmc, which switches the Verilog front-end to recognize SystemVerilog.


src/ebmc/transition_system.cpp
```diff
@@ -221,6 +222,10 @@ int get_transition_system(
   if(cmdline.isset('I'))
     config.verilog.include_paths = cmdline.get_values('I');
 
+  // do --system-verilog
+  if(cmdline.isset("system-verilog"))
+    verilog_languaget::force_systemverilog = true;
```
tautschnig, line 227: We have `languaget::set_language_options`, which seemingly isn't used in the `hw-cbmc` code base (while the cbmc code base does use it). So I think this global variable can be avoided by instead putting information in `options` and making sure that `set_language_options` is called.

src/ebmc/transition_system.cpp
```diff
@@ -21,6 +22,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #include <langapi/mode.h>
 #include <trans-word-level/show_module_hierarchy.h>
 #include <trans-word-level/show_modules.h>
+#include <verilog/verilog_language.h>
```
tautschnig, line 25: That's not actually needed any longer, is it?

tautschnig (approved):

PR #411, opened 2024-03-11
SystemVerilog: structs

kroening:
This adds support for SystemVerilog struct types.


src/verilog/verilog_typecheck_base.cpp
```diff
@@ -182,6 +182,15 @@ std::size_t verilog_typecheck_baset::get_width(const typet &type)
     return (array_size(to_array_type(type)) * element_width).to_ulong();
   }
 
+  if(type.id() == ID_struct)
+  {
+    // add them up
+    mp_integer sum = 0;
+    for(auto &component : to_struct_type(type).components())
+      sum += get_width(component.type());
+    return sum.to_ulong();
```
tautschnig, line 191: Why use `mp_integer` when the final result uses an uncheck conversion to `std::size_t`?

tautschnig (approved):

src/verilog/verilog_typecheck_base.cpp
kroening: Fixed by #417!

PR #412, opened 2024-03-15
Verilog: distinguish packed and unpacked arrays

kroening:
This adds two new IDs to distinguish packed and unpacked arrays in the Verilog parse tree.


src/verilog/verilog_typecheck_type.cpp
```diff
@@ -16,6 +16,79 @@ Author: Daniel Kroening, kroening@kroening.com
...
+
+  typet array_subtype;
+
+  // may need to go recursive
+  if(src_subtype.is_nil())
+    array_subtype = element_type;
+  else
+    array_subtype = array_type(src_subtype, element_type);
+
+  const exprt final_size_expr = from_integer(size, natural_typet());
```
tautschnig, line 82: In #407 the type was changed to `integer_typet`. (And this is now a git conflict that requires resolving.)

PR #417, opened 2024-03-16
Verilog: `get_width` now returns `mp_integer`

kroening:
This avoids checking for overflows when adding up widths.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -266,8 +266,8 @@ exprt verilog_typecheck_exprt::convert_expr_rec(exprt expr)
       throw errort().with_location(expr.source_location())
         << "concatenation expected to have at least one operand";
     }
-    
-    unsigned width=0;
+
+    mp_integer width = 0;
     bool has_verilogbv=false;
 
     Forall_operands(it, expr)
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -2288,16 +2290,16 @@ exprt verilog_typecheck_exprt::convert_trinary_expr(ternary_exprt expr)
...
+        0, unsignedbv_typet{numeric_cast_v<unsigned>(padding_width)});
+      auto new_type = unsignedbv_typet{
+        numeric_cast_v<unsigned>(get_width(op0.type()) + padding_width)};
       expr.op0() = concatenation_exprt(padding, expr.op0(), new_type);
     }
 
     // Part-select expressions are unsigned, even if the
     // entire expression is selected!
-    auto expr_type = unsignedbv_typet((op1 - op2).to_ulong() + 1);
+    auto expr_type = unsignedbv_typet{numeric_cast_v<unsigned>(op1 - op2 + 1)};
```
tautschnig, line 2302: Here and above: is it really ok to use the platform-dependent `unsigned` type?

kroening: Changed to `std::size_t` -- however, I think I really want to change the constructor of the bitvector types to `mp_integer`. Client-code should not have to deal with this.

tautschnig: Yes, indeed.

tautschnig (approved):

PR #425, opened 2024-03-21
Verilog: separate ID for explicit typecast expressions

kroening:
This is to avoid any confusion with automatically generated, implicit typecasts.


tautschnig:
This is failing CI checks. Might be worth adding `-p` to the `test.pl` invocation so that we can actually see in more detail what is going wrong. (The changes seem sane, but presumably we elsewhere introduce an `ID_typecast` expression that is now no longer type checked.)

kroening:
> The changes seem sane, but presumably we elsewhere introduce an `ID_typecast` expression that is now no longer type checked.

Yes, now fixed.

PR #429, opened 2024-03-25
BMC: encode properties using timeframe/handle pair

kroening:
The property encoding interface is changed to return a timeframe/handle pair, instead of just a condition per timeframe.

This allows determining the appropriate size of the counterexample trace.

### DIFF
--- a/src/trans-word-level/property.cpp
+++ b/src/trans-word-level/property.cpp
@@ -70,14 +70,17 @@ Function: property
 
 \*******************************************************************/
 
-void property(
+std::map<std::size_t, exprt::operandst> property_obligations(
   const exprt &property_expr,
-  exprt::operandst &prop_handles,
   message_handlert &message_handler,
   decision_proceduret &solver,
   std::size_t no_timeframes,
   const namespacet &ns)
 {
+  PRECONDITION(no_timeframes > 0);
+
+  std::map<std::size_t, exprt::operandst> obligations;
+
   messaget message(message_handler);
 
   // Initial state only property?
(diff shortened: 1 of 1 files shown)


src/trans-word-level/property.cpp
```diff
@@ -72,7 +72,7 @@ Function: property
 
 void property(
   const exprt &property_expr,
-  exprt::operandst &prop_handles,
+  std::map<std::size_t, exprt::operandst> &obligations,
```
tautschnig, line 75: Shouldn't this actually become the return value?

tautschnig (approved):

src/trans-word-level/property.cpp
kroening: Yes, will do.

src/trans-word-level/property.cpp
```diff
@@ -115,7 +115,45 @@ void property(
       instantiate(p, c, no_timeframes, ns);
 
     auto handle = solver.handle(tmp);
-    prop_handles.push_back(std::move(handle));
+    obligations[c].push_back(std::move(handle));
```
tautschnig, line 118: I guess this isn't really a hot path, but this could safely be written as
````c++
obligations.insert(c, {solver.handle(tmp)});
````
as each index is unique. I'm actually no longer sure why this needs to be `exprt::operandst` as value type when it can only ever be a single `exprt`?

File: src/trans-word-level/property.cpp [hunk 3]
```diff
@@ -87,14 +88,13 @@ void property(
     property_expr.id() == ID_sva_nexttime ||
     property_expr.id() == ID_sva_s_nexttime)
   {
-    prop_handles.resize(no_timeframes, true_exprt());
     if(no_timeframes > 0)
```
tautschnig, line 91: Not really for this PR, but still wondering: is `no_timeframes == 0` actually a legitimate case at all? Should this perhaps be ruled out via a `PRECONDITION`?

src/trans-word-level/property.cpp
kroening: Give me a sec, there will be more than one expression per index, and they'll come separately.

src/trans-word-level/property.cpp
kroening: Yes, I'll make it a PRECONDITION. BMC without any timeframes is not a thing.

PR #430, opened 2024-04-02
ebmc: normalize properties

kroening:
This introduces a pass during which the given property is 'normalized', using heuristic rewrites, with the goal to reduce the burden of a large number of redundant case-splits in each backend.

### DIFF
--- a/src/ebmc/ebmc_properties.cpp
+++ b/src/ebmc/ebmc_properties.cpp
@@ -11,6 +11,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #include <langapi/language.h>
 #include <langapi/language_util.h>
 #include <langapi/mode.h>
+#include <temporal-logic/normalize_property.h>
 #include <verilog/sva_expr.h>
 
 #include "ebmc_error.h"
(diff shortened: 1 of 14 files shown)


src/ebmc/ebmc_properties.cpp
```diff
@@ -70,7 +71,9 @@ ebmc_propertiest ebmc_propertiest::from_transition_system(
       else
         properties.properties.back().name = symbol.pretty_name;
 
-      properties.properties.back().expr = symbol.value;
+      properties.properties.back().original_expr = symbol.value;
+      properties.properties.back().normalized_expr =
+        normalize_property(symbol.value);
```
tautschnig, line 76: I think we'll eventually want a non-default constructor for `propertyt`.

src/temporal-logic/normalize_property.h
```diff
@@ -0,0 +1,16 @@
...
+Author: Daniel Kroening, dkr@amazon.com
+
+\*******************************************************************/
+
+#ifndef CPROVER_TEMPORAL_LOGIC_NORMALIZE_PROPERTY_H
+#define CPROVER_TEMPORAL_LOGIC_NORMALIZE_PROPERTY_H
+
+#include <util/expr.h>
+
+exprt normalize_property(exprt);
```
tautschnig, line 14: Could you please add comments to describe what the normal form is that is being established?

tautschnig (approved):

src/temporal-logic/normalize_property.h
kroening: done

PR #431, opened 2024-04-02
BMC: add support for LTL X operator

kroening:
This adds support for the LTL `X` operator to the BMC engine.


src/trans-word-level/property.cpp
```diff
@@ -70,51 +72,77 @@ Function: property
...
-  }
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+std::map<std::size_t, exprt::operandst> property_obligations(
```
tautschnig, line 136: Should probably always have been `static`

PR #432, opened 2024-04-02
Verilog: ranged always properties

kroening:
This adds support for `always` property expressions with a given range.


src/trans-word-level/property.cpp
```diff
@@ -75,12 +79,12 @@ Function: property_obligations_rec
 static void property_obligations_rec(
   const exprt &property_expr,
   decision_proceduret &solver,
-  std::size_t current,
-  std::size_t no_timeframes,
+  mp_integer current,
+  mp_integer no_timeframes,
```
tautschnig, line 83: Given that `mp_integer` is non-trivial, how about passing those as const references?

src/trans-word-level/property.cpp
```diff
@@ -108,11 +112,51 @@ static void property_obligations_rec(
...
+    {
+      auto to_opt = numeric_cast<mp_integer>(range.op1());
+      if(!to_opt.has_value())
+        throw ebmc_errort() << "failed to convert SVA always to index";
+      to = *to_opt;
+    }
+
+    for(mp_integer c = from; c <= to; ++c)
+    {
+      if(c >= 0 && c < no_timeframes)
```
tautschnig, line 153: A couple of (related) ideas:
* No need to declare `from` before it can be initialised.
* Taking this one step further, how about: `mp_integer from = std::max(0, *from_opt);`
* Use `to = std::min(no_timeframes - 1, *to_opt);`
These together will avoid unnecessary increments of `c` and permit removing the `if` inside this `for` loop.

src/trans-word-level/property.cpp
```diff
@@ -133,13 +177,13 @@ Function: property_obligations
 
 \*******************************************************************/
 
-static std::map<std::size_t, exprt::operandst> property_obligations(
+static std::map<mp_integer, exprt::operandst> property_obligations(
   const exprt &property_expr,
   decision_proceduret &solver,
-  std::size_t no_timeframes,
+  mp_integer no_timeframes,
```
tautschnig, line 183: As above: `const mp_integer &`?

src/verilog/sva_expr.h
```diff
@@ -170,6 +170,129 @@ static inline sva_always_exprt &to_sva_always_expr(exprt &expr)
   return static_cast<sva_always_exprt &>(expr);
 }
 
+class sva_ranged_always_exprt : public binary_predicate_exprt
+{
+public:
+  explicit sva_ranged_always_exprt(binary_exprt range, exprt op)
```
tautschnig, line 176: No need for `explicit`

src/verilog/sva_expr.h
```diff
@@ -170,6 +170,129 @@ static inline sva_always_exprt &to_sva_always_expr(exprt &expr)
...
+{
+  PRECONDITION(expr.id() == ID_sva_ranged_always);
+  sva_ranged_always_exprt::check(expr, validation_modet::INVARIANT);
+  return static_cast<sva_ranged_always_exprt &>(expr);
+}
+
+class sva_s_always_exprt : public binary_predicate_exprt
+{
+public:
+  explicit sva_s_always_exprt(binary_exprt range, exprt op)
```
tautschnig, line 238: No need for `explicit`

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1592,6 +1592,16 @@ void verilog_typecheck_exprt::implicit_typecast(
...
+#if 0
+  else if(src_type.id() == ID_natural)
+  {
+    if(dest_type.id()==ID_integer)
+    {
+      expr = typecast_exprt{expr, dest_type};
+      return;
+    }
+  }
+#endif
```
tautschnig, line 1604: Could a comment please be included to explain why this is defined away?

tautschnig (approved):

src/trans-word-level/property.cpp
kroening: done

src/trans-word-level/property.cpp
kroening: done

src/trans-word-level/property.cpp
kroening: done

src/verilog/verilog_typecheck_expr.cpp
kroening: removed the `#if 0`

PR #445, opened 2024-04-19
Verilog: fix for file in subdirectory with local include

kroening:
This adds a KNOWNBUG test for the case where a Verilog file is in a subdirectory and a local include is in the same subdirectory.


tautschnig:
Requires a rebase to resolve conflicts.

tautschnig (approved):

PR #452, opened 2024-04-21
SMV: set pretty names for top-level identifiers

kroening:
Identifiers in module main now no longer carry the prefix `main::var::`.


tautschnig:
Rebase required to resolve conflicts.

kroening:
done

src/smvlang/expr2smv.h
```diff
@@ -8,6 +8,5 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <util/expr.h>
 
-bool expr2smv(const exprt &expr, std::string &code);
-bool type2smv(const typet &type, std::string &code);
-
+bool expr2smv(const exprt &, std::string &code, const namespacet &);
+bool type2smv(const typet &, std::string &code, const namespacet &);
```
tautschnig, line 12: Not for this PR, but rather asking about potential future work: Is `bool` really the appropriate return type? Shouldn't places that `return true` really be using `UNIMPLEMENTED` instead? Then the return type could instead be `std::string`.

src/smvlang/smv_language.cpp
```diff
@@ -168,7 +172,9 @@ void smv_languaget::show_parse(std::ostream &out, message_handlert &)
       if(it->second.type.id()=="submodule")
       {
         std::string msg;
-        type2smv(it->second.type, msg);
+        symbol_tablet symbol_table;
+        namespacet ns(symbol_table);
```
tautschnig, line 176: Do we really not have a symbol table or namespace available when invoking `show_parse`?

tautschnig (approved):

src/smvlang/expr2smv.h
kroening: Another option would be to send anything that that code doesn't know how to format to `format_expr`.

src/smvlang/smv_language.cpp
kroening: Yes, this is pre typechecking.

PR #461, opened 2024-04-28
SMV netlist output now includes temporal logic properties

kroening:
The conversion to netlist now includes the properties, which are then reproduced in the SMV output.

### DIFF
--- a/regression/ebmc/BDD/BDD_SVA1.desc
+++ b/regression/ebmc/BDD/BDD_SVA1.desc
@@ -6,7 +6,7 @@ BDD_SVA1.sv
 ^\[top\.property\.p0\] always top\.my_bit: PROVED$
 ^\[top\.property\.p1\] always top\.my_bit: PROVED$
 ^\[top\.property\.p2\] always \(top\.counter == 3 \|-> \(nexttime top\.counter == 4\)\): FAILURE: property not supported by BDD engine$
-^\[top\.property\.p3\] always \(top\.counter == 3 \|=> top\.counter == 4\): PROVED$
+^\[top\.property\.p3\] always \(top\.counter == 3 \|=> top\.counter == 4\): FAILURE: property not supported by BDD engine$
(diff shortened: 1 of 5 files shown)


src/trans-netlist/trans_to_netlist.cpp
```diff
@@ -303,6 +308,62 @@ void convert_trans_to_netlistt::operator()(const irep_idt &module)
...
+              return to_sva_s_eventually_expr(expr).op();
+            else
+              PRECONDITION(false);
+          };
+
+          dest.properties.emplace(id, netlistt::GFpt{convert(get_psi(phi))});
+        }
+        else
+        {
+          // unsupported
```
tautschnig, line 361: Should this be `UNIMPLEMENTED;`?

regression/ebmc/BDD/BDD_SVA1.desc
```diff
@@ -6,7 +6,7 @@ BDD_SVA1.sv
 ^\[top\.property\.p0\] always top\.my_bit: PROVED$
 ^\[top\.property\.p1\] always top\.my_bit: PROVED$
 ^\[top\.property\.p2\] always \(top\.counter == 3 \|-> \(nexttime top\.counter == 4\)\): FAILURE: property not supported by BDD engine$
-^\[top\.property\.p3\] always \(top\.counter == 3 \|=> top\.counter == 4\): PROVED$
+^\[top\.property\.p3\] always \(top\.counter == 3 \|=> top\.counter == 4\): FAILURE: property not supported by BDD engine$
```
tautschnig, line 9: How does "netlist output" (as the commit message suggests) result in a change to property checking results?

src/trans-netlist/trans_to_netlist.cpp
kroening: No, I'd prefer this to produce a nice user-facing message in the status output, as opposed to a stack trace.

regression/ebmc/BDD/BDD_SVA1.desc
kroening: This goes as follows: the symbol table to setlist conversion now looks at the properties, and this uncovered a bug in the logic for what is supported by the net-list BMC.

PR #476, opened 2024-05-01
Verilog: create symbols for SVA assumptions

kroening:
Consistently with assertions and cover properties, the Verilog frontend now creates symbols for assumptions.  The engines are changed to ignore these.


src/verilog/verilog_typecheck.cpp
```diff
@@ -1016,51 +1016,46 @@ void verilog_typecheckt::convert_assert_assume_cover(
...
 
-    symbol.module = module_identifier;
-    symbol.value = nil_exprt{}; // set by synthesis
-    symbol.name = full_identifier;
-    symbol.is_property = true;
-    symbol.location = module_item.find_source_location();
-    symbol.pretty_name = strip_verilog_prefix(full_identifier);
+  symbol.module = module_identifier;
+  symbol.value = nil_exprt{}; // set by synthesis
+  symbol.name = full_identifier;
```
tautschnig, line 1048: It needs to be `symbolt symbol{full_identifier, bool_typet{}, mode};` and then `symbol.base_name = base_name;`

tautschnig (changes requested):

tautschnig (approved):

PR #477, opened 2024-05-01
netlist output now includes AIG constraints

kroening:
The output of `--show-netlist` now includes the AIG constraints.

### DIFF
--- a/src/trans-netlist/netlist.cpp
+++ b/src/trans-netlist/netlist.cpp
@@ -56,21 +56,32 @@ void netlistt::print(std::ostream &out) const
 
   out << "Initial state: " << '\n';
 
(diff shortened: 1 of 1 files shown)


src/trans-netlist/netlist.cpp
```diff
@@ -65,6 +65,17 @@ void netlistt::print(std::ostream &out) const
 
   out << '\n';
 
+  out << "State invariant constraints: " << '\n';
+
+  for(unsigned i = 0; i < constraints.size(); i++)
```
tautschnig, line 70: `for(const auto &constraint : constraints)` unless I'm missing something (or at least `std::size_t` instead of `unsigned`).

tautschnig (approved):

PR #478, opened 2024-05-01
state invariants are now converted into AIG state constraints

kroening:
The generation of netlists from transition system now uses the state constraint facility of the netlist data structure for state constraints, as opposed to adding everything as a transition constraint.

This fixes cases where analyzers over-approximate the model semantics.

### DIFF
--- a/src/trans-netlist/trans_to_netlist.cpp
+++ b/src/trans-netlist/trans_to_netlist.cpp
@@ -11,6 +11,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/arith_tools.h>
 #include <util/bitvector_expr.h>
 #include <util/ebmc_util.h>
+#include <util/expr_util.h>
 #include <util/mathematical_expr.h>
 #include <util/namespace.h>
 #include <util/std_expr.h>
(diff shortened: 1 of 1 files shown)


src/trans-netlist/trans_to_netlist.cpp
```diff
@@ -142,6 +146,26 @@ class convert_trans_to_netlistt:public messaget
...
+static bool has_next_symbol(const exprt &expr)
+{
+  auto begin = expr.depth_cbegin(), end = expr.depth_cend();
+  return std::find_if(begin, end, [](const exprt &expr) {
+           return expr.id() == ID_next_symbol;
+         }) != end;
+}
+
+/*******************************************************************\
+
```
tautschnig, line 168: Use `has_subexpr(expr, ID_next_symbol)` from `expr_utils.{h,cpp}` (which does exactly the above).

src/trans-netlist/trans_to_netlist.cpp
```diff
@@ -412,7 +442,11 @@ void convert_trans_to_netlistt::convert_constraints(propt &prop)
   {
     literalt l=
       instantiate_convert(prop, dest.var_map, *it, ns, get_message_handler());
-    transition_constraints.push_back(l);
+
+    if(has_next_symbol(*it))
```
tautschnig, line 446: See above: change this to `if(has_subexpr(*it, ID_next_symbol))`.

tautschnig (approved):

PR #493, opened 2024-05-09
SystemVerilog: $past

kroening:
This adds support for SystemVerilog `$past` expressions to the front-end and to the word-level BMC back-end.


src/verilog/verilog_expr.h
```diff
@@ -1946,4 +1946,45 @@ to_verilog_implicit_typecast_expr(exprt &expr)
...
+{
+  verilog_past_exprt::check(expr);
+  return static_cast<const verilog_past_exprt &>(expr);
+}
+
+inline verilog_past_exprt &to_verilog_past_expr(exprt &expr)
+{
+  verilog_past_exprt::check(expr);
+  return static_cast<verilog_past_exprt &>(expr);
+}
```
tautschnig, line 1988: These should check that `expr.id() == ID_verilog_past`. (Are we perhaps missing this sort of check in some other place where this was initially copied from?)

tautschnig (changes requested):

src/verilog/verilog_expr.h
kroening: Fixed. Will look at the others.

tautschnig (approved):

PR #494, opened 2024-05-09
ebmc: generate a witness that a given candidate isn't a ranking function

kroening:
This records the satisfying assignment when checking a ranking function and outputs it as a two-step trace, to enable debugging the candidate ranking function.

### DIFF
--- a/src/ebmc/ranking_function.cpp
+++ b/src/ebmc/ranking_function.cpp
@@ -114,7 +114,9 @@ int do_ranking_function(
     solver_factory,
     message_handler);
 
-  if(result.is_true())
+  property.witness_trace = std::move(result.second);
+
+  if(result.first.is_true())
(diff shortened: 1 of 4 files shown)


src/ebmc/ranking_function.cpp
```diff
@@ -134,8 +135,11 @@ tvt is_ranking_function(
   const exprt &property,
   const exprt &ranking_function,
   const ebmc_solver_factoryt &solver_factory,
+  std::optional<trans_tracet> &witness_trace,
```
tautschnig, line 138: Not a very big deal, but how about returning `std::pair<tvt, std::optional<trans_tracet>>` (or perhaps create a `struct ranking_function_resultt`)?

tautschnig (approved):

PR #497, opened 2024-05-11
EBMC: give error message when unsupported property is passed to IC3

kroening:
The IC3 engine now reports an appropriate error message when given an unsupported property.

### DIFF
--- a/src/ic3/m1ain.cc
+++ b/src/ic3/m1ain.cc
@@ -12,9 +12,15 @@ Author: Eugene Goldberg, eu.goldberg@gmail.com
 #include <util/ui_message.h>
 
 #include <ebmc/ebmc_properties.h>
+#include <ebmc/report_results.h>
 
 #include <trans-netlist/netlist.h>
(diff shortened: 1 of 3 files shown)


src/ic3/m1ain.cc
```diff
@@ -87,6 +113,30 @@ int ic3_enginet::operator()()
...
+
+      number_of_properties++;
+    }
+
+    if(number_of_properties == 0)
+    {
+      namespacet ns(transition_system.symbol_table);
+      report_results(cmdline, properties, ns, message.get_message_handler());
+      return 10;
+    }
```
tautschnig, line 139: Can you please explain why "10" is the appropriate return value when there were no supported properties? What if there were no properties at all in the first place?

tautschnig (approved):

src/ic3/m1ain.cc
kroening: Will extract into a method in a separate PR, which will hopefully give us consistency.

PR #499, opened 2024-05-12
introduce ebmc_propertiest::exit_code

kroening:
This consolidates the logic for determining the tool exit code in one place, to achieve consistency.

### DIFF
--- a/src/ic3/m1ain.cc
+++ b/src/ic3/m1ain.cc
@@ -135,7 +135,7 @@ int ic3_enginet::operator()()
     {
       namespacet ns(transition_system.symbol_table);
       report_results(cmdline, properties, ns, message.get_message_handler());
-      return 10;
+      return properties.exit_code();
     }
   }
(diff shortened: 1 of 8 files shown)


src/ic3/m1ain.cc
```diff
@@ -135,7 +135,7 @@ int ic3_enginet::operator()()
     {
       namespacet ns(transition_system.symbol_table);
       report_results(cmdline, properties, ns, message.get_message_handler());
-      return 10;
+      return properties.exit_code();
```
tautschnig, line 138: I’m still confused by number-of-properties-equals-zero - should that have been not-equal-to-zero?

tautschnig (approved):

src/ic3/m1ain.cc
kroening: No; the IC3 code right now doesn't use the standard reporting. Will fix.

PR #502, opened 2024-05-12
SMV netlists: SMV identifier syntax

kroening:
SMV identifiers may start with `$`, `_`, `a`-`z`, `A`-`Z`, `0`-`9`.  In addition, `#`, `-`, and `.`  are allowed as subsequent characters.


tautschnig (approved):
Might be worth adding a test.

PR #517, opened 2024-05-21
random traces: set random initial state

kroening:
Instead of relying on the user to initialize the state, we use random constraints.

### DIFF
--- a/regression/ebmc/random-traces/initial_state1.desc
+++ b/regression/ebmc/random-traces/initial_state1.desc
@@ -1,0 +1,10 @@
+CORE
+initial_state1.v
+--random-trace --trace-steps 1 --numbered-trace --random-seed 0
+^m\.x@0 = 'hBF9059EA$
+^m\.y@0 = 123$
+^m\.x@1 = 'hBF9059EB$
+^m\.y@1 = 124$
+^EXIT=0$
+^SIGNAL=0$
+--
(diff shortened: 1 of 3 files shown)


src/ebmc/random_traces.cpp
```diff
@@ -495,6 +579,35 @@ std::vector<exprt> random_tracest::random_input_constraints(
...
+  const symbolst &state_variables)
+{
+  std::vector<exprt> result;
+
+  for(auto &symbol : state_variables)
+  {
+    auto symbol_in_timeframe = instantiate(symbol, 0, 1, ns);
+    auto constraint =
+      equal_exprt(symbol_in_timeframe, random_value(symbol.type()));
+    result.push_back(constraint);
```
tautschnig, line 603: Nit pick: could use `std::move(constraint)`

regression/ebmc/random-traces/initial_state1.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+initial_state1.v
+--random-trace --trace-steps 1 --numbered-trace
```
tautschnig, line 3: I'd suggest to either explicitly add `--random-seed 0` as command line option or add a comment at the bottom that `0` is the default random seed. Else it is not that obvious that random traces will always produce those "random" initial states.

tautschnig (approved):

PR #518, opened 2024-05-21
BMC: include macros in trace

kroening:
This adds the values of macro symbols (`parameter`, `localparam`, etc.) to the BMC trace.

### DIFF
--- a/src/trans-word-level/trans_trace_word_level.cpp
+++ b/src/trans-word-level/trans_trace_word_level.cpp
@@ -58,20 +58,33 @@ trans_tracet compute_trans_trace(
          symbol.type.id()!=ID_module &&
          symbol.type.id()!=ID_module_instance)
       {
-        exprt indexed_symbol_expr(ID_symbol, symbol.type);
+        if(symbol.is_macro)
+        {
+          if(symbol.value.is_constant())
+          {
+            trans_tracet::statet::assignmentt assignment;
+            assignment.rhs = symbol.value;
+            assignment.lhs = symbol.symbol_expr();
+            state.assignments.push_back(assignment);
+          }
+        }
+        else
+        {
+          exprt indexed_symbol_expr(ID_symbol, symbol.type);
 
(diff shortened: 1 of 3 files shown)


src/trans-word-level/trans_trace_word_level.cpp
```diff
@@ -58,20 +58,33 @@ trans_tracet compute_trans_trace(
...
-        assignment.lhs=symbol.symbol_expr();
-      
-        state.assignments.push_back(assignment);
+        if(symbol.is_macro)
+        {
+          if(symbol.value.is_constant())
+          {
+            trans_tracet::statet::assignmentt assignment;
+            assignment.rhs = symbol.value;
+            assignment.lhs = symbol.symbol_expr();
```
tautschnig, line 67: I think `trans_tracet::statet::assignmentt` should only be constructible with two or three (the source location) arguments, unless use with containers absolutely requires the presence of a default constructor.

src/trans-word-level/trans_trace_word_level.cpp
```diff
@@ -58,20 +58,33 @@ trans_tracet compute_trans_trace(
...
+            state.assignments.push_back(assignment);
+          }
+        }
+        else
+        {
+          exprt indexed_symbol_expr(ID_symbol, symbol.type);
+
+          indexed_symbol_expr.set(
+            ID_identifier, timeframe_identifier(t, symbol.name));
```
tautschnig, line 76: Why does this not use `symbol_exprt`?

tautschnig (approved):

src/trans-word-level/trans_trace_word_level.cpp
kroening: Will do as separate PR.

src/trans-word-level/trans_trace_word_level.cpp
kroening: Will change, as separate PR.

PR #525, opened 2024-05-28
extract `transition_systemt::inputs` and `::state_variables`

kroening:
This extracts to replicated methods that identify the inputs and the state variables of the transition system.

### DIFF
--- a/src/ebmc/transition_system.cpp
+++ b/src/ebmc/transition_system.cpp
@@ -368,3 +368,53 @@ int show_symbol_table(
   return get_transition_system(
     cmdline, message_handler, dummy_transition_system);
 }
+
+std::vector<symbol_exprt> transition_systemt::state_variables() const
+{
+  std::vector<symbol_exprt> state_variables;
+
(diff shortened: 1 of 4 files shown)


src/ebmc/transition_system.cpp
```diff
@@ -368,3 +368,53 @@ int show_symbol_table(
   return get_transition_system(
     cmdline, message_handler, dummy_transition_system);
 }
+
+std::vector<symbol_exprt> transition_systemt::state_variables() const
+{
+  std::vector<symbol_exprt> state_variables;
+
+  const auto &module_symbol = *main_symbol;
```
tautschnig, line 376: How do we know this unconditional dereference is safe? Is the use of a pointer the right thing at all?

tautschnig (approved):

src/ebmc/transition_system.cpp
kroening: Not easy to replace by a reference, but will try.

PR #529, opened 2024-06-02
Operators `##[*]` and `##[+]`

kroening:
This adds the sequence operators `##[*]` and `##[+]`.

PR #532, opened 2024-06-04
Fixup #529: add missing .desc file

kroening:
This adds the `.desc` file that was forgotten in #529.


tautschnig (approved):
My apologies for having missed this while reviewing #529.

PR #539, opened 2024-06-11
CI: ebmc-packages workflow fixes

kroening:
This changes the trigger, in the hope it will get run when creating a release, and adds a CentOS 8 rpm.

### DIFF
--- a/.github/workflows/ebmc-packages.yaml
+++ b/.github/workflows/ebmc-packages.yaml
@@ -1,10 +1,11 @@
 on:
   release:
-    types: [created]
+    types: [published]
 
 name: Upload additional release assets
 jobs:
   ubuntu-22_04-package:
+    name: Ubuntu 22.04
     runs-on: ubuntu-22.04
     steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/ebmc-packages.yaml
```diff
@@ -1,10 +1,11 @@
 on:
   release:
-    types: [created]
+    types: [published]
```
tautschnig, line 3: I think it would be better to use tags as the trigger, else you end up publishing a release but don't yet have artifacts in place at the time of publication (and instead hope for them to be created soonish). See https://github.com/model-checking/kani/blob/main/.github/workflows/release.yml#L8-L15 for triggers that appear to work well for us.

tautschnig (approved):

.github/workflows/ebmc-packages.yaml
kroening: Yes, will change once this flow works.

PR #540, opened 2024-06-11
Verilog: add all remaining property_expr rules from 1800-2017

kroening:
This adds all remaining property_expr productions given in IEEE 1800-2017:
* strong
* weak
* case


regression/verilog/SVA/case1.desc
```diff
@@ -0,0 +1,8 @@
+KNOWNBUG
+case1.sv
+--bound 20
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
```
tautschnig, line 8: Can you please add a comment explaining what is going wrong here?

tautschnig (approved):

regression/verilog/SVA/case1.desc
kroening: Added the implementation.

PR #556, opened 2024-06-18
bump CBMC to version 6

kroening:


tautschnig (approved):

regression/verilog/case/case3.v
```diff
@@ -14,9 +14,9 @@ module main(input clk, x, y);
 
   always @(posedge clk)
     casez (cnt1)
-      {10'b0z, 2'b00}:;
-      {10'b0z, 2'b01}:;
-      {10'b0z, 2'b1z}: out=1;
+      10'b0z00:;
+      10'b0z01:;
+      10'b0z1z: out=1;
```
tautschnig, line 19: I did see the earlier failures, but why is it that changes to the tests are the right fix here?

PR #561, opened 2024-06-19
Verilog: aval/bval lowering of 4-valued logic

kroening:
This introduces the standard aval/bval lowering of Verilog's 4-valued logic into bit vectors.


tautschnig (approved):

src/verilog/aval_bval_encoding.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+bv_typet aval_bval_type(std::size_t width, irep_idt source_type)
+{
+  PRECONDITION(!source_type.empty());
+  auto result = bv_typet{width * 2};
+  result.set(ID_C_verilog_aval_bval, source_type);
+  return result;
+}
+
+bv_typet lower_to_aval_bval(const typet &src)
```
tautschnig, line 25: Shouldn't this be a `bitvector_typet` instead of just `typet`? This will then also avoid the `to_bitvector_type` a few lines further down.

src/verilog/aval_bval_encoding.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+exprt bval(const exprt &src)
+{
+  PRECONDITION(is_aval_bval(src.type()));
+  auto width = aval_bval_width(src.type());
+  return extractbits_exprt{
+    src, from_integer(width, integer_typet()), bv_typet{width}};
+}
+
+exprt adjust_size(const exprt &src, std::size_t dest_width)
```
tautschnig, line 110: Could use `static`

src/verilog/aval_bval_encoding.cpp
```diff
@@ -0,0 +1,153 @@
...
+  else if(dest_width < src_width)
+  {
+    return extractbits_exprt{
+      src, from_integer(0, integer_typet{}), bv_typet{dest_width}};
+  }
+  else
+    return src;
+}
+
+exprt combine_aval_bval(const exprt &aval, const exprt &bval, const typet &dest)
```
tautschnig, line 127: Could use `static`

regression/verilog/case/case1.v
```diff
@@ -1,20 +1,20 @@
 module main(input clk, x, y);
 
   reg [1:0] cnt1;
-  reg z;
+  reg result;
```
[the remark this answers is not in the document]
kroening: Yes, `z` could be read as high-impedance state.

PR #564, opened 2024-06-20
Verilog: lowering for concatenation of aval/bval encoded vectors

kroening:
This adds lowering for the concatenation of aval/bval encoded bit-vectors.

### DIFF
--- a/src/verilog/aval_bval_encoding.cpp
+++ b/src/verilog/aval_bval_encoding.cpp
@@ -93,18 +93,31 @@ constant_exprt lower_to_aval_bval(const constant_exprt &src)
 
 exprt aval(const exprt &src)
 {
-  PRECONDITION(is_aval_bval(src.type()));
-  auto width = aval_bval_width(src.type());
-  return extractbits_exprt{
-    src, from_integer(0, integer_typet()), bv_typet{width}};
+  if(is_aval_bval(src.type()))
+  {
+    auto width = aval_bval_width(src.type());
+    return extractbits_exprt{src, 0, bv_typet{width}};
+  }
+  else
+  {
+    auto width = to_bitvector_type(src.type()).get_width();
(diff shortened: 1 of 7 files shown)


src/verilog/aval_bval_encoding.cpp
```diff
@@ -93,18 +93,31 @@ constant_exprt lower_to_aval_bval(const constant_exprt &src)
...
 exprt bval(const exprt &src)
 {
-  PRECONDITION(is_aval_bval(src.type()));
-  auto width = aval_bval_width(src.type());
-  return extractbits_exprt{
-    src, from_integer(width, integer_typet()), bv_typet{width}};
+  if(is_aval_bval(src.type()))
+  {
+    auto width = aval_bval_width(src.type());
+    return extractbits_exprt{src, 0, bv_typet{width}};
```
tautschnig, line 113: Shouldn't this be `width` instead of `0`? (Can we have a test that catches this?)

src/verilog/aval_bval_encoding.cpp
```diff
@@ -93,18 +93,31 @@ constant_exprt lower_to_aval_bval(const constant_exprt &src)
...
+  }
+  else
+  {
+    // zeros, signaling 0/1
+    auto width = to_bitvector_type(src.type()).get_width();
+    return from_integer(0, bv_typet{width});
```
tautschnig, line 119: Maybe throw in a cbmc upgrade and use the new `bv_typet` methods?

tautschnig (changes requested):

src/verilog/aval_bval_encoding.cpp
kroening: Fixed, test added!

tautschnig (approved):

PR #571, opened 2024-06-22
Property description now taken from front-end

kroening:
This moves the generation of the property description into the front-end, which will enable doing lowering or model instrumentation in the front-end while displaying the property as given in the source language.

### DIFF
--- a/src/ebmc/show_properties.cpp
+++ b/src/ebmc/show_properties.cpp
@@ -33,36 +33,37 @@ void ebmc_baset::show_properties()
 {
   unsigned p_nr=1;
 
+  auto make_xml =
+    [](const ebmc_propertiest::propertyt &p, std::size_t p_nr) -> xmlt {
+    xmlt xml("property");
+    xml.set_attribute("name", id2string(p.name));
+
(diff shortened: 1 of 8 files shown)


src/ebmc/show_properties.cpp
```diff
@@ -29,36 +29,37 @@ void ebmc_baset::show_properties()
 {
   unsigned p_nr=1;
 
+  auto make_xml =
+    [](const ebmc_propertiest::propertyt &p, std::size_t p_nr) -> xmlt {
+    xmlt xml("property");
+    xml.set_attribute("name", id2string(p.name));
+
+    xml.new_element("number").data = std::to_string(p_nr); // will go away
```
tautschnig, line 37: Just curious: what will go away and when?

tautschnig (approved):

src/ebmc/show_properties.cpp
kroening: That's just a moved comment, not new. You can't use the number for anything (we use the property identifier instead), so it's really redundant.

PR #575, opened 2024-06-23
SystemVerilog: concurrent assertion items go into module namespace

kroening:
Concurrent assertion items (assert/assume/cover) that are module items with a block label go into the module name space, not into a separate "property" name space.

Assertion items without block item now get a label based on the kind of the assertion item (assert/assume/cover).

### DIFF
--- a/regression/ebmc/BDD/BDD2.desc
+++ b/regression/ebmc/BDD/BDD2.desc
@@ -3,6 +3,6 @@ BDD2.sv
 --module main --bdd
 ^EXIT=10$
 ^SIGNAL=0$
-^\[main.property.my_prop1\] always main.counter < 199: PROVED$
-^\[main.property.my_prop2\] always main.counter < 198: REFUTED$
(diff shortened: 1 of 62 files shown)


tautschnig:
> Concurrent assertion items (assert/assume/cover) that are module items with a block label go into the module name space, not into a separate "property" name space.
> 
> Assertion items without block item now get a label based on the kind of the assertion item (assert/assume/cover).

I can see that the commit does what the above message says, but _why_ is this the right thing to do?

kroening:
> I can see that the commit does what the above message says, but _why_ is this the right thing to do?

One argument in favour is that it appears to be what other tools are doing (e.g., XCelium uses the same scheme for assertions with label, and `__assert_.`number for assertions without).

kroening:
I of course wish that the IEEE standard be clear about it, but couldn't find anything.

tautschnig (approved):

PR #578, opened 2024-06-24
Verilog: assignment pattern expressions

kroening:
This adds support for struct and array literals to the SystemVerilog frontend by implementing the `assignment_pattern_expression` production rules.


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1614,6 +1628,34 @@ void verilog_typecheck_exprt::implicit_typecast(
...
+      {
+        // rec. call
+        implicit_typecast(expr.operands()[i], components[i].type());
+      }
+
+      // turn into struct expression
+      expr.id(ID_struct);
+      expr.type() = dest_type;
+      return;
+    }
```
tautschnig, line 1657: The earlier comment said these could become structs or arrays -- are the latter just not yet implemented?

kroening: Array case added.

tautschnig (approved):

PR #585, opened 2024-06-30
Verilog: generate property description prior to expression synthesis

kroening:
This generates the property description from the expression prior to synthesis, which is closer to the Verilog source code.

### DIFF
--- a/regression/verilog/SVA/named_property1.desc
+++ b/regression/verilog/SVA/named_property1.desc
@@ -1,7 +1,7 @@
 CORE
 named_property1.sv
 --bound 0
-^\[main\.assert\.1\] always main\.x == 10: PROVED up to bound 0$
+^\[main\.assert\.1\] always main\.x_is_ten: PROVED up to bound 0$
 ^EXIT=0$
 ^SIGNAL=0$
 --
--- a/src/verilog/expr2verilog.cpp
+++ b/src/verilog/expr2verilog.cpp
@@ -647,6 +647,81 @@ std::string expr2verilogt::convert_index(
 
 /*******************************************************************\
 
+Function: expr2verilogt::convert_non_indexed_part_select
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+std::string expr2verilogt::convert_non_indexed_part_select(
+  const verilog_non_indexed_part_select_exprt &src,
+  verilog_precedencet precedence)
+{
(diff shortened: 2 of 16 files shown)


src/verilog/expr2verilog.cpp
```diff
@@ -863,6 +938,26 @@ std::string expr2verilogt::convert_next_symbol(
...
+
+ Purpose:
+
+\*******************************************************************/
+
+std::string expr2verilogt::convert_hierarchical_identifier(
+  const hierarchical_identifier_exprt &src,
+  unsigned &precedence)
+{
+  precedence = 22;
```
tautschnig, line 955: Could you please add a comment for this magic number?

src/verilog/expr2verilog.cpp
```diff
@@ -1061,6 +1156,18 @@ std::string expr2verilogt::convert(
...
+    src.id() == ID_verilog_indexed_part_select_plus ||
+    src.id() == ID_verilog_indexed_part_select_minus)
+  {
+    return convert_indexed_part_select(
+      to_verilog_indexed_part_select_plus_or_minus_expr(src), precedence = 18);
+  }
+
+  else if(src.id() == ID_verilog_non_indexed_part_select)
+    return convert_non_indexed_part_select(
+      to_verilog_non_indexed_part_select_expr(src), precedence = 18);
```
tautschnig, line 1169: Could you please add comments for those magic numbers?

tautschnig (approved):

regression/verilog/SVA/named_property1.desc
```diff
@@ -1,7 +1,7 @@
 CORE
 named_property1.sv
 --bound 0
-^\[main\.assert\.1\] always main\.x == 10: PROVED up to bound 0$
+^\[main\.assert\.1\] always main\.x_is_ten: PROVED up to bound 0$
```
tautschnig, line 4: I would have expected `x_is_ten` to make it into the property identifier, but maybe that's a misconception on my part?

kroening: Ah, no, there is a some terminology issue here: The identifier is for assertions, not Verilog properties. Verilog properties may or may not end up as an assertion, assumption, cover statement, etc.

Verilog properties do not get checked by default.

src/verilog/expr2verilog.cpp
kroening: #591

PR #586, opened 2024-06-30
AIG engine: fix for new symbols for fresh inputs

kroening:
When generating the AIG, the conversion adds new symbols for the next-state value if none is given.  These new symbols now have a mode, and a fresh identifier is generated for each bit, to ensure that the type (bool) is consistent with the variable map.

### DIFF
--- a/src/trans-netlist/trans_to_netlist.cpp
+++ b/src/trans-netlist/trans_to_netlist.cpp
@@ -59,6 +59,8 @@ protected:
   netlistt &dest;
   
   literalt new_input();
+  std::size_t input_counter = 0;
+  irep_idt mode;
 
   class rhs_entryt
   {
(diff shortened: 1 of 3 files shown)


src/trans-netlist/trans_to_netlist.cpp
```diff
@@ -168,6 +170,7 @@ literalt convert_trans_to_netlistt::new_input()
     symbol.type=bool_typet();
     symbol.is_input=true;
     symbol.base_name="input";
+    symbol.mode = mode;
```
tautschnig, line 173: Perhaps change the lines above to use the `symbolt symbol{id, bool_typet{}, mode};` constructor.

tautschnig (approved):

PR #573, opened 2024-06-22
JSON output for the modules

kroening:
This adds `--json-modules`, to get a JSON array with all the modules.

PR #588, opened 2024-07-04
introduce output_filet

kroening:
This introduces a new class `output_filet`, which

* given a filename, opens the file for writing, and provides an `std::ostream` for it; if this fails, an appropriate exception is thrown.

* given "-", provides a pointer to `std::cout`

### DIFF
--- a/src/ebmc/Makefile
+++ b/src/ebmc/Makefile
@@ -20,6 +20,7 @@ SRC = \
       live_signal.cpp \
       main.cpp \
       neural_liveness.cpp \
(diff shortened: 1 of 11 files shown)


tautschnig (approved):
Wouldn't this also be useful in the CBMC code base and, hence, should live over there instead?

tautschnig:
The code now merged via #573 should now also be revisited.

kroening:
> Wouldn't this also be useful in the CBMC code base and, hence, should live over there instead?

Yes, will do.

PR #607, opened 2024-07-25
Add KNOWNBUG test for SMV ranges

kroening:


regression/ebmc/range_type/range_type6.desc
```diff
@@ -0,0 +1,8 @@
+KNOWNBUG
+range_type6.smv
+--bound 10
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+The result does not match the expected result.
```
tautschnig, line 8: Will this one be fixed by doing a CBMC submodule update?

kroening: Yes!

PR #614, opened 2024-07-31
add word-level BMC encoding for SVA sequences

kroening:
This adds a BMC encoding for SVA sequences by translating an SVA sequence expression into a list of match point timeframe numbers and match conditions.


src/trans-word-level/instantiate_word_level.cpp
```diff
@@ -175,24 +169,131 @@ wl_instantiatet::instantiate_rec(exprt expr, const mp_integer &t) const
...
+    for(auto &lhs_match_point : lhs_match_points)
+    {
+      // The RHS of the non-overlapped implication starts one timeframe later
+      auto t_rhs = expr.id() == ID_sva_non_overlapped_implication
+                     ? lhs_match_point.first + 1
+                     : lhs_match_point.first;
+
+      // Do we exceed the bound? Make it 'true'
+      if(t_rhs >= no_timeframes)
+        return {{no_timeframes - 1, true_exprt()}};
```
tautschnig, line 205: Could `no_timeframes` be zero? (And: does it matter if it is, i.e., is it a problem if the result includes a negative number?)

src/trans-word-level/instantiate_word_level.cpp
```diff
@@ -175,24 +169,131 @@ wl_instantiatet::instantiate_rec(exprt expr, const mp_integer &t) const
...
+  {
+    return {t, timeframe_symbol(t, to_symbol_expr(std::move(expr)))};
+  }
+  else if(is_SVA_sequence(expr))
+  {
+    // sequence expressions -- these may have multiple potential
+    // match points, and evaluate to true if any of them matches
+    const auto match_points = instantiate_sequence(expr, t);
+    exprt::operandst disjuncts;
+    mp_integer max = t;
```
tautschnig, line 288: `t` isn't necessarily one of the values of `match_point.first`, is that a problem?

tautschnig (approved):

src/trans-word-level/instantiate_word_level.cpp
```diff
@@ -175,24 +169,131 @@ wl_instantiatet::instantiate_rec(exprt expr, const mp_integer &t) const
...
+  else if(expr.id() == ID_symbol)
+  {
+    return {t, timeframe_symbol(t, to_symbol_expr(std::move(expr)))};
+  }
+  else if(is_SVA_sequence(expr))
+  {
+    // sequence expressions -- these may have multiple potential
+    // match points, and evaluate to true if any of them matches
+    const auto match_points = instantiate_sequence(expr, t);
+    exprt::operandst disjuncts;
```
tautschnig, line 287: Nit pick: I'd throw in `disjuncts.reserve(match_points.size());`.

src/trans-word-level/instantiate_word_level.cpp
kroening: It's `bound + 1`, where `bound` is unsigned, so shouldn't. But could add a DATA_INVARIANT.

src/trans-word-level/instantiate_word_level.cpp
kroening: All the `match_point.first` are `>=t`, and hence, `max` is going to be one of them unless there is no match point at all.

PR #616, opened 2024-08-01
SystemVerilog: sequence repetition operators

kroening:
This adds the SVA sequence repitition operators.


tautschnig (approved):
Would it be worth adding a KNOWNBUG test (which you might have handy for exercising the parser)?

ShashankVM:
Thanks Dr. Daniel for this fix, I'd like to use it, may I know if this change will get reflected if I do a git pull?

PR #618, opened 2024-08-04
SystemVerilog: SVA abort properties

kroening:
This adds SVA abort properties (IEEE 1800-2017 16.12.14).


src/temporal-logic/normalize_property.h
```diff
@@ -20,6 +20,10 @@ Author: Daniel Kroening, dkr@amazon.com
 /// sva_s_nexttime φ --> Xφ
 /// sva_if --> ? :
 /// a sva_disable_iff b --> a ∨ b
+/// a sva_accept_on b --> a ∨ b
+/// a sva_reject_on b --> ¬a ∧ b
+/// a sva_sync_accept_on b --> a ∨ b
```
tautschnig, line 25: Are there really three different ways that all just express disjunction?

tautschnig (approved):
A few more tests would be nice.

src/temporal-logic/normalize_property.h
kroening: Yes. They differ slightly if you are building a simulator.

PR #619, opened 2024-08-05
SystemVerilog: $countones

kroening:
This adds the SystemVerilog system function `$countones` (IEEE 1800-2017 20.9).

### DIFF
--- a/regression/verilog/system-functions/countones2.desc
+++ b/regression/verilog/system-functions/countones2.desc
@@ -1,0 +1,7 @@
+CORE
+countones2.sv
+--bound 50
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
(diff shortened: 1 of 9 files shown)


regression/verilog/system-functions/countones2.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+countones2.sv
+--bound 10
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
```
tautschnig, line 7: Why is this (still) a `KNOWNBUG`, i.e., what aspect of this test is unexpectedly not working?

kroening: Now works!

tautschnig (approved):

PR #621, opened 2024-08-07
bump version number to 5.1

kroening:


tautschnig (approved):
May I suggest that we start maintaining a CHANGELOG of sorts? Perhaps with the help of GitHub's "Generate release notes" capability?

PR #627, opened 2024-08-08
Verilog: set defines on command line

kroening:
This allows setting SystemVerilog preprocessor defines on the command line with `-D`.


tautschnig (approved):

tautschnig:
Shouldn't there also be a test that actually seeks to parse the result after preprocessing (with command-line set defines)?

PR #629, opened 2024-08-10
Verilog: constant folding for reduction operators

kroening:


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1545,13 +1545,68 @@ exprt verilog_typecheck_exprt::elaborate_constant_expression(exprt expr)
...
+      }
+      else if(type.id() == ID_signedbv)
+      {
+        return from_integer(-1, type);
+      }
+      else if(type.id() == ID_bool)
+        return true_exprt{};
+      else
+        PRECONDITION(false);
+    };
```
tautschnig, line 1563: Wouldn't `bv_typet::all_ones_expr` do the trick?

tautschnig (approved):

src/verilog/verilog_typecheck_expr.cpp
kroening: Not sure if the simplifier can cast that? Will check.

PR #633, opened 2024-08-14
`verilog_synthesist::current_value` can be const

kroening:
This method can be const by replacing the `[]` operator on a map that is used by `.find(...)`.


tautschnig (approved):
Approving, but needs a rebase to resolve conflicts.

PR #634, opened 2024-08-14
Verilog: separate synthesis for LHS expressions

kroening:
The left-hand side of assignments requires special-case synthesis; the symbol to be assigned must not be replaced by its value, but any array indices must be.

### DIFF
--- a/src/verilog/verilog_synthesis.cpp
+++ b/src/verilog/verilog_synthesis.cpp
@@ -440,6 +440,80 @@ verilog_synthesist::synthesis_constant(const exprt &expr)
 
 /*******************************************************************\
 
+Function: verilog_synthesist::synth_lhs_expr
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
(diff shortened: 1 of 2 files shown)


src/verilog/verilog_synthesis.cpp
```diff
@@ -405,6 +405,69 @@ exprt verilog_synthesist::synth_expr(exprt expr, symbol_statet symbol_state)
...
+  {
+    return expr; // leave as is
+  }
+  else if(expr.id() == ID_concatenation)
+  {
+    for(auto &op : expr.operands())
+      op = synth_lhs_expr(op);
+
+    return expr;
+  }
```
tautschnig, line 430: What are the semantics of a concatenation as an lvalue?

tautschnig (commented):
Needs a rebase.

src/verilog/verilog_synthesis.cpp
kroening: Piece-wise assignment to each component.

tautschnig (approved):

PR #637, opened 2024-08-14
Verilog: fix synthesis for assignment to indexed part select

kroening:


regression/verilog/part-select/indexed-part-select5.desc
```diff
@@ -0,0 +1,9 @@
+KNOWNBUG
+indexed-part-select5.sv
+--bound 0
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+This yields invalid SMT-LIB ("invalid extract application").
```
tautschnig, line 9: For that to materialise, wouldn't also some `--smt` command line be required?

kroening: We do run all tests both with the flattening backend and with Z3.

tautschnig (approved):

PR #640, opened 2024-08-21
Verilog: fix assignment for module output ports

kroening:
This flips the assignments generated by synthesis for module output ports to read

  `c := port`

for a connected signal `c`, instead of `port := c`.


tautschnig (approved):
Is there a test for this?

kroening:
> Is there a test for this?

I am afraid not -- the order only becomes a problem with the changes in #639.

PR #641, opened 2024-08-24
Verilog: remove verilog_modulet

kroening:
This removes one of the four representations for Verilog modules.


src/verilog/verilog_parse_tree.cpp
```diff
@@ -35,18 +35,19 @@ void verilog_parse_treet::create_module(
...
 
-  verilog_modulet new_module;
+  verilog_module_sourcet new_module;
 
-  new_module.name=name.id();
-  new_module.parameter_port_list.swap(parameter_port_list);
-  new_module.ports.swap(ports);
-  new_module.location=((const exprt &)module_keyword).source_location();
-  new_module.module_items.swap(module_items);
+  new_module.base_name(name.id());
```
tautschnig, line 40: Could `verilog_module_sourcet` be altered so that it at least requires a base name upon construction?

tautschnig (approved):

PR #644, opened 2024-08-25
Verilog: add attributes into parse tree

kroening:
Verilog allows attributes (key-value pairs) in various parts of the grammar. These are now added to the generated parse tree.


src/verilog/parser.y
```diff
@@ -858,13 +881,20 @@ module_item_brace:
...
-	| attribute_instance_brace module_instantiation { $$=$2; }
-        | attribute_instance_brace smv_using { $$ = $2; }
-        | attribute_instance_brace smv_assume { $$ = $2; }
-	| attribute_instance_brace module_common_item { $$=$2; }
+	  attribute_instance_brace parameter_override
+		{ add_attributes($2, $1); $$=$2; }
+	| attribute_instance_brace gate_instantiation
+		{ add_attributes($2, $1); $$=$2; }
+	// | attribute_instance_brace udp_instantiation
+	//	{ add_attributes($2, $1); $$=$2; }
```
tautschnig, line 889: Why was this always commented out (and still is)?

src/verilog/parser.y
```diff
@@ -907,10 +942,15 @@ class_item_brace:
 
 class_item:
 //	  attribute_instance_brace class_property
+//		{ add_attributes($2, $1); $$=$2; }
 //	| attribute_instance_brace class_method
+//		{ add_attributes($2, $1); $$=$2; }
 //	| attribute_instance_brace class_constraint
+//		{ add_attributes($2, $1); $$=$2; }
```
tautschnig, line 949: Why was/is this commented out?

src/verilog/parser.y
```diff
@@ -873,16 +903,21 @@ module_or_generate_item_declaration:
...
 /*
 	  module_or_generate_item
-	| attribute_instance_brace parameter_declaration { $$=$2; }
-	// | attribute_instance_brace local_parameter_declaration { $$=$2; }
+	| attribute_instance_brace parameter_declaration
+	//	{ add_attributes($2, $1); $$=$2; }
+	// | attribute_instance_brace local_parameter_declaration
+	//	{ add_attributes($2, $1); $$=$2; }
 	;
 */
```
tautschnig, line 922: Why was/is this commented out?

src/verilog/parser.y
```diff
@@ -157,6 +157,25 @@ static void add_as_subtype(typet &dest, typet &what)
...
+
+ Purpose:
+
+\*******************************************************************/
+
+static void add_attributes(YYSTYPE &dest, YYSTYPE &attributes)
+{
+  PRECONDITION(stack_expr(attributes).id() == ID_verilog_attributes);
+  if(!stack_expr(attributes).get_sub().empty())
+    addswap(dest, ID_verilog_attributes, attributes);
```
tautschnig, line 174: Would it perhaps be better to use `ID_C_verilog_attributes`? They seem to be ignored during type checking at this point, so I am guessing they don't affect the semantics?

src/verilog/parser.y
kroening: Classes are simply not done yet.

src/verilog/parser.y
kroening: These are not in the 1800-2017 grammar -- will clean out.

src/verilog/parser.y
kroening: I don't believe anything with attributes ever gets compared for equality, so it wouldn't make any difference.

The attributes seem to be entirely tool-specific, so there may well be some that have/alter semantics.

I am planning to use them to configure reasoning engines per assertion, but it's not obvious that this is the best mechanism.

src/verilog/parser.y
tautschnig: Perhaps you can document this in source via a comment?

src/verilog/parser.y
tautschnig: Perhaps this should be documented?

tautschnig (approved):

src/verilog/parser.y
kroening: Done

src/verilog/parser.y
kroening: Done

PR #648, opened 2024-08-27
Verilog: `create_module` now returns `verilog_module_sourcet`

kroening:
Instead of storing the module via a side-effect, `create_module` now returns the `verilog_module_sourcet`.

This enables nested modules.


tautschnig (approved):
> This enables nested modules.

This suggests to me there should be a new test added?

PR #651, opened 2024-08-30
introduce `property_checker_resultt`

kroening:
This introduces a type for the result from an engine, as a replacement for `int`.

### DIFF
--- a/src/ebmc/property_checker.h
+++ b/src/ebmc/property_checker.h
@@ -12,6 +12,13 @@ Author: Daniel Kroening, dkr@amazon.com
 #include "ebmc_properties.h"
 #include "transition_system.h"
 
+enum class property_checker_resultt
+{
+  VERIFICATION_RESULT,
+  SUCCESS,
+  ERROR
+};
+
 int property_checker(
   const cmdlinet &,
   transition_systemt &,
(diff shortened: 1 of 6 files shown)


src/ebmc/property_checker.h
```diff
@@ -12,6 +12,13 @@ Author: Daniel Kroening, dkr@amazon.com
 #include "ebmc_properties.h"
 #include "transition_system.h"
 
+enum class property_checker_resultt
+{
+  REPORT_RESULT,
```
tautschnig, line 17: I am struggling to understand what this one is about? Or, rather, why is "SUCCESS" not a "result?"

tautschnig:
Also, is there a case for re-using `src/goto-checker/properties.h`?

src/ebmc/property_checker.h
kroening: It's not a verification result -- rename REPORT_RESULT -> REPORT_VERIFICATION_RESULT?

kroening:
> Also, is there a case for re-using `src/goto-checker/properties.h`?

It will move into this direction; but there simply aren't any goto programs.

src/ebmc/property_checker.h
kroening: Or simply `VERIFICATION_RESULT`?

tautschnig: `VERIFICATION_RESULT` works for me.

kroening: done

tautschnig (approved):

PR #660, opened 2024-09-01
Verilog: fix for signed constants

kroening:
Verilog allows constants to be marked as signed (e.g., 'sb1).  When using base 2, 8 and 16, these need to be sign extended to 32 bits unless the number of bits is given.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -16,6 +16,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/prefix.h>
 #include <util/simplify_expr.h>
 #include <util/std_expr.h>
+#include <util/string2int.h>
 
 #include "expr2verilog.h"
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1223,14 +1223,14 @@ exprt verilog_typecheck_exprt::convert_constant(constant_exprt expr)
...
-  unsigned bits=0;
+  std::size_t bits = 0;
   bool bits_given=false;
 
   if(pos!=std::string::npos) // size given?
   {
     if(rest[0]!='\'')
     {
-      bits=atoi(rest.c_str());
+      bits = atol(rest.c_str());
```
tautschnig, line 1233: This is probably safe because of the lexer rules, but I would feel better if this used one of the `safe_string2...` functions instead of `atol`.

tautschnig (approved):

PR #662, opened 2024-09-01
Verilog: constant output now includes signedness and width

kroening:
The conversion of constants to Verilog source now includes the signedness and width when outputting hex or binary literals.

### DIFF
--- a/src/verilog/expr2verilog.cpp
+++ b/src/verilog/expr2verilog.cpp
@@ -1020,14 +1020,20 @@ std::string expr2verilogt::convert_constant(
     auto i = numeric_cast_v<mp_integer>(src);
 
     if(i>=256)
-      dest="'h"+integer2string(i, 16);
+    {
+      dest = std::to_string(width);
+      dest += "'h";
+      if(type.id() == ID_signedbv)
+        dest += 's';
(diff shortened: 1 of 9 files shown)


src/verilog/expr2verilog.cpp
```diff
@@ -1020,14 +1020,20 @@ std::string expr2verilogt::convert_constant(
     auto i = numeric_cast_v<mp_integer>(src);
 
     if(i>=256)
-      dest="'h"+integer2string(i, 16);
+    {
+      dest = std::to_string(width);
+      dest = "'h";
```
tautschnig, line 1025: The second write will just replace the first value - should this be `+=` instead? Also, this suggests a test case is missing here?

tautschnig (changes requested):

src/verilog/expr2verilog.cpp
kroening: Indeed -- fixed, and test added.

tautschnig: Err, sorry, one of the previously existing test now needs to have its pattern adjusted (rightly so).

tautschnig (approved):
Approving, just needs more fixes to existing tests.

PR #671, opened 2024-09-05
Verilog: `{}` and `{{}}` do not need parentheses

kroening:
This changes the precedence of concatenation and replication when generating Verilog from expressions such that no additional parentheses are generated.

### DIFF
--- a/src/verilog/expr2verilog_class.h
+++ b/src/verilog/expr2verilog_class.h
@@ -19,29 +19,29 @@ class sva_ranged_predicate_exprt;
 
 // Precedences (higher means binds more strongly).
 // Follows Table 11-2 in IEEE 1800-2017.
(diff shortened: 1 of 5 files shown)


src/verilog/expr2verilog_class.h
```diff
@@ -19,29 +19,29 @@ class sva_ranged_predicate_exprt;
 
 // Precedences (higher means binds more strongly).
 // Follows Table 11-2 in IEEE 1800-2017.
-//
+// We deviate from the table for the precedence of concatenation
+// and replication, which act like parenteses.
```
tautschnig, line 23:
```suggestion
// and replication, which act like parentheses.
```

tautschnig (approved):

PR #672, opened 2024-09-06
introduce `expr2verilogt::resultt`

kroening:
This introduces a struct to return the pair of string and precedence in expr2verilog, instead of just returning the string, and assigning the precedence to a reference parameter.


tautschnig (approved):
Can we please also have this in expr2ct? :-)

PR #684, opened 2024-09-08
SystemVerilog: grammar for distribution weights

kroening:
This adds the 1800-2017 rules for distribution weights.


tautschnig (approved):
Should this come with a test?

PR #687, opened 2024-09-11
Fix fragment of SVA supported by word-level BMC

kroening:


tautschnig (approved):
Now that there is just one commit left in this PR I am wondering whether there should be test just for what this final commit changes?

kroening:
> Now that there is just one commit left in this PR I am wondering whether there should be test just for what this final commit changes?

Test added!

PR #691, opened 2024-09-14
SVA: implement ranged `s_eventually` operator

kroening:
This adds a BMC encoding for the ranged variant of the `s_eventually` operator.


src/trans-word-level/instantiate_word_level.cpp
```diff
@@ -428,6 +429,45 @@ wl_instantiatet::instantiate_rec(exprt expr, const mp_integer &t) const
     DATA_INVARIANT(no_timeframes != 0, "must have timeframe");
     return {no_timeframes - 1, conjunction(conjuncts)};
   }
+  else if(expr.id() == ID_sva_ranged_s_eventually)
+  {
```
tautschnig, line 433: Not that this change makes the decisive difference, but: `instantiate_rec` has become very very long and could do with breaking into smaller functions. Perhaps this opportunity could be taken to at least move this case to a function of its own?

src/trans-word-level/instantiate_word_level.cpp
```diff
@@ -428,6 +429,45 @@ wl_instantiatet::instantiate_rec(exprt expr, const mp_integer &t) const
...
+    auto from = t + std::max(mp_integer{0}, *from_opt);
+
+    mp_integer to;
+
+    if(upper.id() == ID_infinity)
+    {
+      throw ebmc_errort()
+        << "failed to convert SVA s_eventually to index (infinity)";
+    }
+    else
```
tautschnig, line 451: I believe this check could be done earlier so that `to` could use RAII and also its handling would be symmetric to that of `from`.

tautschnig (approved):

src/trans-word-level/instantiate_word_level.cpp
kroening: Will do. What bugs me most is the duplication between instantiate_word_level.cpp and property.cpp.

PR #697, opened 2024-09-16
Verilog: rename convert_type method to elaborate_type

kroening:
This renames the method that elaborates types, given that it follows the elaboration procedure in the Verilog standard.


tautschnig (approved):
Requires a rebase.

PR #705, opened 2024-09-19
SMV: reject CTL in LTLSPEC, and LTL in CTLSPEC

kroening:
The SMV typechecker now rejects CTL when checking LTLSPEC, and LTL when checking CTLSPEC.


tautschnig (approved):
clang-format wants to be made happy.

PR #708, opened 2024-09-19
expr2smv: enum for precedence

kroening:
This replaces the use of `unsigned` and assorted integer literals for the expression precedence by an enum type.

### DIFF
--- a/src/smvlang/expr2smv.cpp
+++ b/src/smvlang/expr2smv.cpp
@@ -21,80 +21,98 @@ public:
   {
   }
 
-  bool convert_nondet_choice(const exprt &src, std::string &dest, unsigned precedence);
+protected:
+  // In NuSMV 2.6., ! (not)  has a high precedence (above ::), whereas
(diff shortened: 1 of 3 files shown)


src/smvlang/expr2smv.cpp
```diff
@@ -21,33 +21,75 @@ class expr2smvt
...
+    INDEX = 16,  // [ ] , [ : ]
+    CONCAT = 15, // ::
+    UMINUS = 14, // - (unary minus)
+    MULT = 13,   // * / mod
+    PLUS = 12,   // + -
+    SHIFT = 11,  // << >>
+    UNION = 10,  // union
+    IN = 9,      // in
+    REL = 8,     // = != < > <= >=
+    TEMP = 7,    // AX, AF, etc.
```
tautschnig, line 40: Perhaps spell out "TEMPORAL" on this occasion? When reading the code I found myself thinking "wait, is there some sort of _temporary_ precedence?"

tautschnig (approved):

PR #712, opened 2024-09-21
BMC: support full LTL

kroening:
Full LTL is supported by the word-level BMC engine.

### DIFF
--- a/src/trans-word-level/property.cpp
+++ b/src/trans-word-level/property.cpp
@@ -40,44 +40,7 @@ Function: bmc_supports_LTL_property
 
 bool bmc_supports_LTL_property(const exprt &expr)
 {
-  // We support
-  // * formulas that contain no temporal operator besides X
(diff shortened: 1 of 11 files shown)


src/trans-word-level/property.cpp
```diff
@@ -40,44 +40,7 @@ Function: bmc_supports_LTL_property
...
-  else if(expr.id() == ID_and)
-  {
-    for(auto &op : expr.operands())
-      if(!bmc_supports_LTL_property(op))
-        return false;
-    return true;
-  }
-  else
-    return false;
+  return true;
```
tautschnig, line 43: Should this function just be removed?

tautschnig (approved):

src/trans-word-level/property.cpp
kroening: Look at the pretty symmetry of `bmc_supports_property(...)`!

PR #714, opened 2024-09-22
Verilog: `convert_range` now returns range

kroening:
The `convert_range` method now returns the range, as opposed to assigning to two reference parameters.

### DIFF
--- a/src/verilog/verilog_elaborate_type.cpp
+++ b/src/verilog/verilog_elaborate_type.cpp
@@ -44,12 +44,10 @@ array_typet verilog_typecheck_exprt::convert_unpacked_array_type(
   if(range_expr.is_not_nil())
   {
     // these may be negative
-    mp_integer msb, lsb;
-    convert_range(range_expr, msb, lsb);
-    big_endian = (lsb > msb);
-    size = (big_endian ? lsb - msb : msb - lsb) + 1;
-    DATA_INVARIANT(size >= 0, "array size must not be negative");
-    offset = big_endian ? msb : lsb;
+    auto range = convert_range(range_expr);
+    big_endian = (range.lsb > range.msb);
+    size = range.length();
+    offset = range.smallest_index();
   }
   else if(size_expr.is_not_nil())
   {
@@ -96,16 +94,15 @@ typet verilog_typecheck_exprt::convert_packed_array_type(
 {
   PRECONDITION(src.id() == ID_verilog_packed_array);
 
-  const exprt &range = static_cast<const exprt &>(src.find(ID_range));
+  const exprt &range_expr = static_cast<const exprt &>(src.find(ID_range));
   const auto &source_location = src.source_location();
 
-  mp_integer msb, lsb;
-  convert_range(range, msb, lsb);
+  auto range = convert_range(range_expr);
 
-  bool big_endian = (lsb > msb);
+  bool big_endian = (range.lsb > range.msb);
 
-  mp_integer width = (big_endian ? lsb - msb : msb - lsb) + 1;
-  mp_integer offset = big_endian ? msb : lsb;
+  mp_integer width = range.length();
+  mp_integer offset = range.smallest_index();
 
   // let's look at the subtype
   const auto subtype =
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_elaborate_type.cpp
```diff
@@ -44,12 +44,10 @@ array_typet verilog_typecheck_exprt::convert_unpacked_array_type(
...
-    mp_integer msb, lsb;
-    convert_range(range_expr, msb, lsb);
-    big_endian = (lsb > msb);
-    size = (big_endian ? lsb - msb : msb - lsb) + 1;
-    DATA_INVARIANT(size >= 0, "array size must not be negative");
-    offset = big_endian ? msb : lsb;
+    auto range = convert_range(range_expr);
+    big_endian = (range.lsb > range.msb);
+    size = range.length();
+    offset = big_endian ? range.msb : range.lsb;
```
tautschnig, line 50:
```suggestion
    offset = range.smallest_index();
```

src/verilog/verilog_interfaces.cpp
```diff
@@ -163,14 +163,14 @@ void verilog_typecheckt::interface_inst(
...
+  const exprt &range_expr = static_cast<const exprt &>(op.find(ID_range));
 
-  mp_integer msb, lsb;
-  
-  if(range.is_nil() || range.id().empty())
-    msb=lsb=0;
+  ranget range;
+
+  if(range_expr.is_nil() || range_expr.id().empty())
+    range.msb = range.lsb = 0;
```
tautschnig, line 171: Shouldn't there be a default constructor of `ranget` that does this?

src/verilog/verilog_typecheck.cpp
```diff
@@ -203,18 +203,19 @@ Function: verilog_typecheckt::typecheck_builtin_port_connections
...
-  exprt &range=static_cast<exprt &>(inst.add(ID_range));
+  exprt &range_expr = static_cast<exprt &>(inst.add(ID_range));
 
-  mp_integer msb, lsb;
-  
-  if(range.is_nil() || range.id()=="")
-    msb=lsb=0;
+  ranget range;
+
+  if(range_expr.is_nil() || range_expr.id() == "")
```
tautschnig, line 210: Should use `.empty()` or `== irep_idt()`, see also comment above.

src/verilog/verilog_typecheck.cpp
```diff
@@ -203,18 +203,19 @@ Function: verilog_typecheckt::typecheck_builtin_port_connections
...
+    range.msb = range.lsb = 0;
   else
-    convert_range(range, msb, lsb);
+    range = convert_range(range_expr);
+
+  if(range.lsb > range.msb)
+    std::swap(range.lsb, range.msb);
```
tautschnig, line 216: Why is this here (and just here) the right thing to do?

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -2101,19 +2101,20 @@ Function: verilog_typecheck_exprt::convert_range
...
   {
     throw errort().with_location(range.source_location())
       << "range expected to have two operands";
   }
 
-  msb = convert_integer_constant_expression(to_binary_expr(range).op0());
-  lsb = convert_integer_constant_expression(to_binary_expr(range).op1());
+  ranget result;
+  result.msb = convert_integer_constant_expression(to_binary_expr(range).op0());
+  result.lsb = convert_integer_constant_expression(to_binary_expr(range).op1());
```
tautschnig, line 2115: There should be a 2-argument constructor for `ranget`.

tautschnig (approved):

src/verilog/verilog_interfaces.cpp
kroening: Considered it, but would you remember whether msb is the first argument, or lsb?

kroening: I'll go for msb, lsb.

src/verilog/verilog_typecheck.cpp
kroening: It's probably broken. I'll investigate.

PR #722, opened 2024-09-23
Homebrew formula for ebmc 5.1

kroening:


tautschnig (approved):
Should release automation create this file upon every release? In case it helps: https://github.com/diffblue/homebrew-cbmc/blob/main/.github/workflows/add-formula.yml does this for CBMC.

kroening:
> Should release automation create this file upon every release? In case it helps: https://github.com/diffblue/homebrew-cbmc/blob/main/.github/workflows/add-formula.yml does this for CBMC.

Yes, this should be automated.

PR #727, opened 2024-09-24
Verilog: aval/bval encoding for `!`

kroening:
This adds the aval/bval encoding for Verilog negation.

### DIFF
--- a/regression/verilog/expressions/negation1.desc
+++ b/regression/verilog/expressions/negation1.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE broken-smt-backend
 negation1.sv
 --bound 0
 ^EXIT=0$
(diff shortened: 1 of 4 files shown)


regression/verilog/expressions/negation1.desc
```diff
@@ -1,9 +1,8 @@
-KNOWNBUG
+CORE broken-smt-backend
```
tautschnig, line 1: Just curious: which feature are we missing in the SMT encoding for this test?

tautschnig (approved):

regression/verilog/expressions/negation1.desc
kroening: verilog_case_equality

PR #733, opened 2024-09-25
BMC: fix encoding of `p U q`

kroening:
p U q can be refuted using a ¬q loop.


src/hw_cbmc_irep_ids.h
```diff
@@ -9,7 +9,9 @@ IREP_ID_ONE(EX)
 IREP_ID_ONE(EU)
 IREP_ID_ONE(ER)
 IREP_ID_ONE(U)
+IREP_ID_ONE(weak_U)
 IREP_ID_ONE(R)
+IREP_ID_ONE(strong_R)
```
tautschnig, line 14: Where is that used?

kroening: You'd get that by negating weak until, and we don't have a frontend that generates it. Will drop.

PR #738, opened 2024-09-29
SystemVerilog: conversion from/to packed union

kroening:
This adds conversions from/to packed unions to/from vector types.

### DIFF
--- a/src/verilog/verilog_synthesis.cpp
+++ b/src/verilog/verilog_synthesis.cpp
@@ -11,6 +11,7 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <util/bitvector_expr.h>
 #include <util/bitvector_types.h>
+#include <util/c_types.h>
 #include <util/ebmc_util.h>
 #include <util/expr_util.h>
 #include <util/identifier.h>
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_synthesis.cpp
```diff
@@ -113,6 +114,20 @@ exprt verilog_synthesist::from_bitvector(
 
     return struct_exprt{std::move(field_values), struct_type};
   }
+  else if(dest.id() == ID_union)
+  {
+    // We use the first field of the union.
```
tautschnig, line 119: Just curious: does Verilog require that all fields have the same width? If not, could there be any problems if the first field wasn't sufficiently wide?

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -2146,6 +2146,32 @@ typet verilog_typecheck_exprt::enum_decay(const typet &type)
...
+{
+  // 1800-2017 7.3.1
+  // Verilog union types decay to a vector type [$bits(t)-1:0] when
+  // used in relational or arithmetic expressions.
+  auto &type = expr.type();
+  if(type.id() == ID_union)
+  {
+    auto new_type =
+      unsignedbv_typet{numeric_cast_v<std::size_t>(get_width(type))};
+    expr = typecast_exprt{expr, new_type};
```
tautschnig, line 2169: Will this reach our back-end? I don't think we'd support casts from a union. I suspect we'd need an extractbits here instead.

src/verilog/verilog_synthesis.cpp
kroening: Yes, in Verilog packed unions all fields must be the same width. Will add comment to say so.

src/verilog/verilog_typecheck_expr.cpp
kroening: Yes, the synthesis stage turns these into a member expression -- will add comment.

tautschnig (approved):

PR #743, opened 2024-10-01
word-level BMC: allow SVA and/or/implies/iff at property level

kroening:


regression/verilog/SVA/sva_iff2.sv
```diff
@@ -0,0 +1,11 @@
+module main(input a, b);
+
+  // p0: assert property ((always a) iff (always a));
+  p1: assert property ((eventually[0:1] a) iff (eventually[0:1] a));
+  // p2: assert property ((s_eventually a) iff (s_eventually a));
+  p3: assert property ((a until b) iff (a until b));
+  // p4: assert property ((a s_until b) iff (a s_until a));
```
tautschnig, line 7: Why are some of those commented out?

tautschnig (approved):

regression/verilog/SVA/sva_iff2.sv
kroening: Still failing -- this is the same problem in `regression/verilog/SVA/eventually4.sv`

PR #750, opened 2024-10-04
Verilog: set type of implicit nets

kroening:
1800 2017 6.10 allows implicit declarations of nets.  The type of these nets is to be derived from the LHS of the assignment or the type of the port connection.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -893,7 +893,7 @@ exprt verilog_typecheck_exprt::convert_nullary_expr(nullary_exprt expr)
   }
   else if(expr.id()==ID_symbol)
   {
-    return convert_symbol(to_symbol_expr(std::move(expr)));
+    return convert_symbol(to_symbol_expr(std::move(expr)), {});
   }
   else if(expr.id()==ID_verilog_star_event)
   {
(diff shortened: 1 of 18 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1023,19 +1025,23 @@ exprt verilog_typecheck_exprt::convert_symbol(symbol_exprt expr)
...
   else
   {
-    throw errort().with_location(expr.source_location())
-      << "unknown identifier " << identifier;
+    if(implicit_net_type.has_value())
+    {
+      implicit_wire(identifier, symbol, implicit_net_type.value());
+
+      warning().source_location = expr.source_location();
+      warning() << "implicit wire " << symbol->display_name() << eom;
```
tautschnig, line 1035: Why does this yield a warning, or, rather: in what way should the input have been modified?

tautschnig (approved):

src/verilog/verilog_typecheck_expr.cpp
tautschnig: If this is covered by the standard (as the commit message suggests), why is such a recommendation warranted?

kroening: ok, will drop

PR #767, opened 2024-10-15
change Fφ BMC encoding

kroening:
This changes the word-level BMC encoding for Fφ properties to remove counterexample traces that exhibit the following:
    
1) a ¬φ loop,
2) followed by one or more φ states.
    
While these traces demonstrate that a counterexample exists (constructed by simply expanding the ¬φ loop), they are not themselves counterexamples to Fφ.

### DIFF
--- a/src/trans-word-level/property.cpp
+++ b/src/trans-word-level/property.cpp
@@ -256,6 +256,19 @@ static obligationst property_obligations_rec(
 
     obligationst obligations;
 
+    // Traces with any φ state from "current" onwards satisfy Fφ
+    exprt::operandst phi_disjuncts;
+
+    phi_disjuncts.reserve(numeric_cast_v<std::size_t>(no_timeframes - current));
+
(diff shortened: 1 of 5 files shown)


src/trans-word-level/property.cpp
```diff
@@ -256,6 +256,17 @@ static obligationst property_obligations_rec(
 
     obligationst obligations;
 
+    // Traces with any φ state from "current" onwards satisfy Fφ
+    exprt::operandst phi_disjuncts;
```
tautschnig, line 260: We should make a habit of using `.reserve` when populating the container via `push_back` a known quantity of times. It'll take an extra type conversion here to go from `mp_integer` to `std::size_t`, but we should avoid the potentially quadratic cost of re-allocating and copying.

tautschnig (approved):

src/trans-word-level/property.cpp
kroening: Done

PR #787, opened 2024-10-23
Verilog: constant folding for indexed part select expressions

### LINKED ISSUE #784: Expression not recognized as constant when using indexed part-select
It seems that a similar problem as in #751 arises if indexed part-select is used instead of part-select.

If I run the following example as `ebmc examples/test_constant_expr.v --systemverilog --top test_constant_expr --bound 10`
```
module test_constant_expr #(parameter x = 8'hff) ();

    test_mod #(.p(x[3 -: 2])) mod ();

endmodule

module test_mod #(parameter p = 1) ();
    
    initial begin
        $display("p = 0x%0h", p);
    end
endmodule
```
I get the following output
```
Parsing examples/test_const_part_select.v
Converting
Type-checking Verilog::test_mod
Type-checking Verilog::test_const_part_select
file examples/test_const_part_select.v line 4: expected constant expression, but got `16'hFFFF[8-4]'
CONVERSION ERROR
```

I ran the example with latest commits on main branch (fb9a9d48) that already includes the fix for #751.

kroening:
This implements constant folding for indexed part select expressions, following the pattern in #766.

Fixes #784.

### DIFF
--- a/src/verilog/verilog_expr.cpp
+++ b/src/verilog/verilog_expr.cpp
@@ -168,3 +168,80 @@ exprt verilog_non_indexed_part_select_exprt::lower() const
 {
   return ::lower(*this);
 }
+
+static exprt
+lower(const verilog_indexed_part_select_plus_or_minus_exprt &part_select)
+{
+  auto get_width = [](const typet &t) -> mp_integer
+  {
+    if(t.id() == ID_bool)
+      return 1;
+
+    if(
+      t.id() == ID_unsignedbv || t.id() == ID_signedbv ||
+      t.id() == ID_verilog_signedbv || t.id() == ID_verilog_unsignedbv)
+    {
+      return to_bitvector_type(t).get_width();
+    }
+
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_expr.cpp
```diff
@@ -168,3 +168,82 @@ exprt verilog_non_indexed_part_select_exprt::lower() const
...
+    if(part_select.id() == ID_verilog_indexed_part_select_plus)
+    {
+      bottom = index_int - src_offset;
+      top = bottom + width - 1;
+    }
+    else // ID_verilog_indexed_part_select_minus
+    {
+      top = index_int - src_offset;
+      bottom = bottom - width + 1;
+    }
```
tautschnig, line 223: `top` isn't actually needed (anymore, thanks to your changes to `extractbits_exprt`).

tautschnig (approved):

PR #793, opened 2024-10-31
BMC: push negation into conjunction

kroening:
This pushes the negation into the conjunction of the property constraint for BMC.  This yields a constraint that is a disjunction, which is CNF-friendly.


tautschnig (approved):
Would you have any numbers to clarify how much of a difference this makes?

kroening:
Closing in favor of #805

PR #796, opened 2024-11-01
Verilog: lowering for four-valued zero-extend

kroening:
This adds a lowering for `zero_extend` expressions for four-valued operands.


kroening:
This has been done in #1476.

PR #805, opened 2024-11-10
BMC with iterative constraint strengthening

kroening:
This implements "iterative constraint strengthening" in the word-level BMC
engine (VMCAI 2009, Query-driven Program Testing), and makes it the default.
The previous, assumption-based method remains available with
`--bmc-with-assumptions`.

When using MiniSat, the benefit on hwmcc08 is barely measurable.  When using
Cadical, the benefit is around 10%.

MiniSat iterative   88.27s
MiniSat asumptions  88.73s
Cadical iterative  103.41s
Cadical asumptions 117.74s

### DIFF
--- a/regression/verilog/functioncall/functioncall2.desc
+++ b/regression/verilog/functioncall/functioncall2.desc
@@ -3,6 +3,6 @@ functioncall2.v
 --module main --bound 1
 ^EXIT=0$
 ^SIGNAL=0$
-^UNSAT: No counterexample found within bound$
+^UNSAT: .*$
 --
 ^warning: ignoring
--- a/src/ebmc/bmc.cpp
+++ b/src/ebmc/bmc.cpp
@@ -17,9 +17,184 @@ Author: Daniel Kroening, dkr@amazon.com
 #include <chrono>
 #include <fstream>
 
+void bmc_with_assumptions(
+  std::size_t bound,
+  const transition_systemt &transition_system,
+  ebmc_propertiest &properties,
+  decision_proceduret &solver,
+  message_handlert &message_handler)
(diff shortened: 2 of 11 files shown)


regression/verilog/functioncall/functioncall2.desc
```diff
@@ -3,6 +3,6 @@ functioncall2.v
 --module main --bound 1
 ^EXIT=0$
 ^SIGNAL=0$
-^UNSAT: No counterexample found within bound$
+^UNSAT: .*$
```
tautschnig, line 6: Can we use this (or any other) test to exercise bmc-with-assumptions?

src/ebmc/bmc.cpp
```diff
@@ -17,9 +17,181 @@ Author: Daniel Kroening, dkr@amazon.com
...
+{
+  messaget message(message_handler);
+  const namespacet ns(transition_system.symbol_table);
+
+  auto trace_found = [&solver](const ebmc_propertiest::propertyt &property)
+  {
+    for(auto &h : property.timeframe_handles)
+      if(solver.get(h).is_false())
+        return true;
+    return false;
```
tautschnig, line 109: How about return solver.get(h).is_false();

tautschnig (approved):

src/ebmc/bmc.cpp
kroening: That won't work -- note the loop.

regression/verilog/functioncall/functioncall2.desc
kroening: Test added

PR #834, opened 2024-11-20
Verilog: lowering for four-valued `->`

kroening:
This adds a lowering for Verilog's `->` implication expression, with four-valued semantics.


regression/verilog/expressions/constants2.sv
```diff
@@ -31,7 +31,7 @@ module main;
   parameter p27 = 1<<<1;
   parameter p28 = 1&&1;
   parameter p29 = 1||1;
-  parameter p30 = 1->1;
+//  parameter p30 = 1->1;
```
tautschnig, line 34: Why is this newly commented out (in this particular commit)?

tautschnig (approved):

regression/verilog/expressions/constants2.sv
kroening: Fixed by rebasing on main, owing to #835

PR #842, opened 2024-11-21
SystemVerilog: type parameter ports

kroening:
This implements grammar for SystemVerilog type parameter ports.


kroening:
This has been done in #1412.

PR #849, opened 2024-11-24
SystemVerilog: KNOWNBUG test for `return` statement

kroening:


regression/verilog/functioncall/return1.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+return1.sv
+--bound 0
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
```
tautschnig, line 7: Would appreciate a sentence on the problem here? Is it also a test that fails to parse?

tautschnig (changes requested):

tautschnig (approved):

PR #859, opened 2024-12-01
BMC: Implement SVA sequence intersect

kroening:


src/trans-word-level/sequence.cpp
```diff
@@ -132,7 +132,31 @@ std::vector<std::pair<mp_integer, exprt>> instantiate_sequence(
   else if(expr.id() == ID_sva_sequence_intersect)
   {
     // IEEE 1800-2017 16.9.6
-    PRECONDITION(false);
+    // - Both the operands shall match.
+    // — The lengths of the two matches of the operand sequences shall be the same.
```
tautschnig, line 136: Where does the code enforce these two preconditions? It seems the code below is missing an else branch?!

kroening: Clarified the comment; this is not a precondition, but the condition for the intersection to match.

src/verilog/sva_expr.h
```diff
@@ -1181,4 +1181,29 @@ to_sva_sequence_consecutive_repetition_expr(exprt &expr)
   return static_cast<sva_sequence_consecutive_repetition_exprt &>(expr);
 }
 
+class sva_sequence_intersect_exprt : public binary_exprt
+{
+public:
+  explicit sva_sequence_intersect_exprt(exprt op0, exprt op1)
```
tautschnig, line 1187: No need for “explicit”

tautschnig (approved):

PR #866, opened 2024-12-03
vlindex: properties and sequences

kroening:
This adds indexing for SystemVerilog named properties and named sequences.


tautschnig (approved):
Would it be possible to add a vlindex run to the RISC-V CI job?

PR #869, opened 2024-12-03
SystemVerilog: retain explicit casts

kroening:
This preserves the explicit casts when typechecking SystemVerilog, which enables outputting them as part of property descriptions.

They are subsequently lowered during synthesis.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -761,7 +761,7 @@ exprt verilog_typecheck_exprt::convert_system_function(
 
   if(identifier=="$signed")
   {
-    // this is a type cast
+    // this is an explicit type cast
     if(arguments.size()!=1)
     {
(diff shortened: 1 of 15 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -809,21 +807,20 @@ exprt verilog_typecheck_exprt::convert_system_function(
...
+      new_type.id(ID_unsignedbv);
+      expr.type() = new_type;
+      return std::move(expr);
     }
     else if(argument.type().id()==ID_bool)
     {
-      typecast_exprt tmp(argument, unsignedbv_typet(1));
-      tmp.add_source_location()=expr.source_location();
-      return std::move(tmp);
+      expr.type() = unsignedbv_typet{1};
```
tautschnig, line 822: Why is this signedbv_typet{2} in the very similar code above?

tautschnig (approved):

PR #875, opened 2024-12-03
SVA: KNOWNBUG test for a recursive property

kroening:


regression/verilog/SVA/recursive_property1.desc
```diff
@@ -0,0 +1,9 @@
+CORE
```
tautschnig, line 1: Should be KNOWNBUG

PR #891, opened 2024-12-20
Verilog: add rounding mode to casts to real/shortreal

kroening:
Casts to `real`/`shortreal` may need to round, hence add the rounding mode during lowering.

### DIFF
--- a/regression/verilog/expressions/cast_to_real1.desc
+++ b/regression/verilog/expressions/cast_to_real1.desc
@@ -1,0 +1,8 @@
+CORE broken-smt-backend
+cast_to_real1.sv
+--bound 0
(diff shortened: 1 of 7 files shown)


tautschnig:
Should there also be a test that newly passes? As is right now, this adds both a KNOWNBUG test and code changes, the effect of the latter not being very clear as far as their impact on tests is concerned.

kroening:
> Should there also be a test that newly passes? As is right now, this adds both a KNOWNBUG test and code changes, the effect of the latter not being very clear as far as their impact on tests is concerned.

I've split up the test.

tautschnig (approved):

PR #896, opened 2024-12-22
EBMC: basic engine selection heuristic

kroening:
This adds a trivial engine selection heuristic, as follows:

1) First, try 1-induction, which may refute (base case) or prove (step case) the property.

2) If still unresolved, try BMC with bound 5.

The heuristic is activated when no engine is specified on the command line.

### DIFF
--- a/src/ebmc/property_checker.cpp
+++ b/src/ebmc/property_checker.cpp
@@ -356,6 +356,59 @@ property_checker_resultt bit_level_bmc(
   }
 }
 
+property_checker_resultt engine_heuristic(
+  const cmdlinet &cmdline,
+  transition_systemt &transition_system,
+  ebmc_propertiest &properties,
+  message_handlert &message_handler)
+{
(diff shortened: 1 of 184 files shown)


src/ebmc/property_checker.cpp
```diff
@@ -356,6 +356,56 @@ property_checker_resultt bit_level_bmc(
...
+    return property_checker_resultt::error();
+  }
+
+  auto solver_factory = ebmc_solver_factory(cmdline);
+
+  if(!properties.has_unknown_property())
+    return property_checker_resultt{properties}; // done
+
+  // First try 1-induction, word-level
+  message.status() << "Attempting 1-induction" << messaget::eom;
```
tautschnig, line 379: How about inserting one more line of output that tells the user that they haven't specified an engine and thereby triggered heuristic (maybe it should actually be called "default"?) selection?

tautschnig (approved):

PR #898, opened 2024-12-23
IC3: better error message when given assumption

kroening:
This improves the error handling when giving an assumption to the IC3 engine.


tautschnig:
Could this one please be rebased as IC3 result reporting changes might break newly added tests!?

tautschnig (approved):

PR #902, opened 2024-12-24
SMV typechecker cleanup

kroening:
1. `smv_ranget` is moved into a header file

2. Use `override` instead of `virtual` for the methods in `smv_typecheckt`.

3. Remove unnecessary declarators on method parameters.

4. Use ranged `for` instead of `Forall_operands` and `forall_operands`.

### DIFF
--- a/src/smvlang/smv_typecheck.cpp
+++ b/src/smvlang/smv_typecheck.cpp
@@ -82,6 +82,7 @@ protected:
   void convert(smv_parse_treet::modulet::itemt &);
   void typecheck(smv_parse_treet::modulet::itemt &);
   void typecheck_expr_rec(exprt &, const typet &, modet);
+  void convert_expr_to(exprt &, const typet &dest);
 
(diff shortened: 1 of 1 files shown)


src/smvlang/smv_range.h
```diff
@@ -0,0 +1,88 @@
...
+  void to_type(typet &dest) const
+  {
+    dest = typet(ID_range);
+    dest.set(ID_from, integer2string(from));
+    dest.set(ID_to, integer2string(to));
+  }
+
+  bool is_bool() const
+  {
+    return from == 0 && to == 1;
```
tautschnig, line 52: While I understand that this is the same behaviour as before: why would from == 1 or to == 0 not be permitted?

tautschnig (approved):

src/smvlang/smv_range.h
kroening: See #903, and will add a precondition to the class.

PR #923, opened 2025-01-06
Engine heuristic: fix for assumptions unsupported by k-induction

### LINKED ISSUE #921: Problem with engine heuristic
The new basic engine heuristic seems to introduce a bug when it is applied to designs containing assumptions not supported by the k-induction engine (temporal property in my case). The assumption seems to be transformed into an assertion.

As an example, if you take ```regression/ebmc/k-induction/ring_buffer.sv```, and you slightly modify the first assumption in order to add a temporal operator as follows, the engine will output ```REFUTED``` instead of ```ASSUMED``` or ```FAILURE: property unsupported by k-induction``` for the first assumption:

```
module ring_buffer(input clk, input read, input write);

  reg [4:0] count=0;
  reg [3:0] readptr=0, writeptr=0;

  always @(posedge clk) begin
    if(read) begin
      readptr++;
      count--;
    end

    if(write) begin
      writeptr++;
      count++;
    end

  end

  wire full=count==16;
  wire empty=count==0;

  assume property (empty |=> !read);
  assume property (full |-> !write);

  p0: assert property (((writeptr-readptr)&'b1111)==count[3:0]);
  p1: assert property (count <= 16);
  p2: assert property (count != 17);

endmodule
```

kroening:
The k-induction engine now correctly reports unsupported assumptions, and is then skipped by the engine selection heuristic.
    
Fixes #921.

### DIFF
--- a/src/trans-word-level/property.cpp
+++ b/src/trans-word-level/property.cpp
@@ -696,6 +696,7 @@ void property(
   auto obligations = property_obligations(property_expr, no_timeframes);
 
   // Map obligations onto timeframes.
+  prop_handles.clear();
   prop_handles.resize(no_timeframes, true_exprt());
   for(auto &obligation_it : obligations.map)
   {
--- a/regression/ebmc/engine-heuristic/basic1.desc
+++ b/regression/ebmc/engine-heuristic/basic1.desc
@@ -1,0 +1,10 @@
+CORE
+basic1.sv
+
+^\[main\.a0\] always not s_eventually !main\.x: ASSUMED$
+^\[main\.p0\] always main\.x: PROVED up to bound 5$
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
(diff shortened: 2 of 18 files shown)


src/trans-word-level/property.cpp
```diff
@@ -696,6 +696,7 @@ void property(
   auto obligations = property_obligations(property_expr, no_timeframes);
 
   // Map obligations onto timeframes.
+  prop_handles.clear();
   prop_handles.resize(no_timeframes, true_exprt());
```
tautschnig, line 700: Since this is now always cleared shouldn't this be returned by value rather than being required as argument?

tautschnig (approved):

src/trans-word-level/property.cpp
kroening: Yes, will action.

kroening: Will be separate PR.

kroening: #934

PR #934, opened 2025-01-16
BMC: return property handles/literals

kroening:
The property handles/literals for the BMC time frames are now returned as return values, as opposed to being returned via a reference.

This removes the ambiguity about what should happen when there are values in those vectors prior to the call.

### DIFF
--- a/src/trans-netlist/unwind_netlist.cpp
+++ b/src/trans-netlist/unwind_netlist.cpp
@@ -141,21 +141,22 @@ Function: unwind_property
 
 \*******************************************************************/
 
-void unwind_property(
+bvt unwind_property(
   const netlistt::propertyt &property,
-  const bmc_mapt &bmc_map,
-  bvt &prop_bv)
+  const bmc_mapt &bmc_map)
 {
   PRECONDITION(std::holds_alternative<netlistt::Gpt>(property));
   auto property_node = std::get<netlistt::Gpt>(property).p;
 
-  prop_bv.resize(bmc_map.timeframe_map.size());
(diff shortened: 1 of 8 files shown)


src/trans-netlist/unwind_netlist.cpp
```diff
@@ -141,21 +141,23 @@ Function: unwind_property
...
   const netlistt::propertyt &property,
-  const bmc_mapt &bmc_map,
-  bvt &prop_bv)
+  const bmc_mapt &bmc_map)
 {
   PRECONDITION(std::holds_alternative<netlistt::Gpt>(property));
   auto property_node = std::get<netlistt::Gpt>(property).p;
 
+  bvt prop_bv;
   prop_bv.resize(bmc_map.timeframe_map.size());
```
tautschnig, line 152:
```suggestion
  bvt prop_bv{bmc_map.timeframe_map.size()};
```
or perhaps use `.reserve` instead of `.resize` and use `.push_back` in the loop below.

src/trans-word-level/property.cpp
```diff
@@ -696,7 +695,7 @@ void property(
   auto obligations = property_obligations(property_expr, no_timeframes);
 
   // Map obligations onto timeframes.
-  prop_handles.clear();
+  exprt::operandst prop_handles;
   prop_handles.resize(no_timeframes, true_exprt());
```
tautschnig, line 699:
```suggestion
  exprt::operandst prop_handles{no_timeframes, true_exprt()};
```

tautschnig (approved):

Issue #931, opened 2025-01-16
CONVERSION ERROR


Fuhj-better:
I don't know why I can't handle expressions like |=> and |->. According to the manual, it should be supported. 
Here is my code:
property p_push_wr_ptr;
    @(posedge clk)
    push && !full |=> wr_ptr == ($past(wr_ptr) + 1);
endproperty
assert property (p_push_wr_ptr); 

Here is the error message:
file fifo.sv line 50: no conversion for binary expression sva_non_overlapped_implication
CONVERSION ERROR
Why does this happen?

Fuhj-better:
> That got broken -- a candidate fix is in [#933](https://github.com/diffblue/hw-cbmc/pull/933), I would be grateful if you could confirm.

Thank you very much for your reply and help. I have encountered a new problem.
Here is my code:
line 16: always_ff @(posedge clk or negedge rst_n) begin

Here is the error message:
root@:~/sv-test/counter# ebmc fifo.sv --top fifo --bound 10
Parsing fifo.sv
Converting
Type-checking Verilog::fifo
file fifo.sv line 16: pos/negedge expected to have Boolean as operand, but got [0:0]
CONVERSION ERROR

I haven't determined yet whether the issue with property ... endproperty has been resolved.

Fuhj-better:
> That got broken -- a candidate fix is in [#933](https://github.com/diffblue/hw-cbmc/pull/933), I would be grateful if you could confirm.这个问题被打破了——候选修复已在 [#933](https://github.com/diffblue/hw-cbmc/pull/933) ，如果您能确认，我将不胜感激。

And when I processed the following assertions, I encountered an error:
verilog复制
assert property (@(posedge clk) (mgmt_st == `MGM_POWERUP) |-> (sdr_init_done == 1'b0));
assert property (@(posedge clk) (mgmt_st == `MGM_ACTIVE) |-> (sdr_init_done == 1'b1));
assert property (@(posedge clk) (mgmt_cmd == `SDR_PRECHARGE) |-> ##[trp_delay:trp_delay] (mgmt_cmd != `SDR_PRECHARGE));
assert property (@(posedge clk) (mgmt_cmd == `SDR_REFRESH) |-> ##[trcar_delay:trcar_delay] (mgmt_cmd != `SDR_REFRESH));
assert property (@(posedge clk) (mgmt_cmd == `SDR_MODE) |-> ##[4:4] (mgmt_cmd != `SDR_MODE));
[...]

PR #936, opened 2025-01-17
SVA: add test with non-constant SVA cycle delays

kroening:
Replicates third example in #931.

### DIFF
--- a/regression/verilog/SVA/cycle_delay1.desc
+++ b/regression/verilog/SVA/cycle_delay1.desc
@@ -1,0 +1,9 @@
+CORE
+cycle_delay1.sv
+
+^file cycle_delay1\.sv line \d+: expected constant expression, but got `main.from'$
+^EXIT=2$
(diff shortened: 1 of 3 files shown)


regression/verilog/SVA/cycle_delay1.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+cycle_delay1.sv
+
+^C string exception : failed to convert sva_cycle_delay offsets$
+^EXIT=6$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+The error message could be better.
```
tautschnig, line 10: While this is true I am worried no one will ever pay attention to it again? Perhaps now is the time to fix it?

tautschnig (approved):

regression/verilog/SVA/cycle_delay1.desc
kroening: Done

PR #949, opened 2025-01-28
Verilog: introduce scope stack

kroening:
This adds a scope stack to the Verilog parser, to enable entering named scopes using the `::` and member operators while parsing.


tautschnig:
It seems the parser changes need more work here: this is failing tests.

kroening:
> It seems the parser changes need more work here: this is failing tests.

Will split into smaller bits.

tautschnig (approved):

PR #968, opened 2025-02-05
SystemVerilog: discover package dependencies in expressions and types

kroening:
Dependencies on a package via the package scope operator `::` are now discovered when used in types or expressions.

### DIFF
--- a/src/verilog/verilog_expr.cpp
+++ b/src/verilog/verilog_expr.cpp
@@ -11,6 +11,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/arith_tools.h>
 #include <util/bitvector_expr.h>
 #include <util/bitvector_types.h>
+#include <util/expr_iterator.h>
 #include <util/mathematical_types.h>
 #include <util/prefix.h>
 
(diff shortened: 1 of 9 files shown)


tautschnig:
Is the change to the cbmc submodule intentional?

kroening:
Fixed

src/verilog/verilog_expr.cpp
```diff
@@ -81,6 +81,34 @@ void verilog_module_sourcet::show(std::ostream &out) const
...
+  {
+    auto &package_scope_expr = to_verilog_package_scope_expr(expr);
+    dest.push_back(
+      verilog_package_identifier(package_scope_expr.package_base_name()));
+  }
+  else
+  {
+    for(auto &op : expr.operands())
+      dependencies_rec(op, dest);
+  }
```
tautschnig, line 99: In the interest of reducing (or: not further increasing) the use of recursion when traversing expressions (which at times is causing stack overflows for us): can we please use an expr iterator here:
````C++
for(const_depth_iteratort it = expr.depth_cbegin(); it != expr.depth_cend(); ++it)
{
  if(it->id() == ID_verilog_package_scope)
  {
    auto &package_scope_expr = to_verilog_package_scope_expr(*it);
    dest.push_back(
      verilog_package_identifier(package_scope_expr.package_base_name()));
  }
}
````

tautschnig (approved):

PR #982, opened 2025-02-09
SMV: do not allow defining identifiers that are already declared

kroening:
`VAR x : ...` followed by `DEFINE x := ...` must be errored.

### DIFF
--- a/regression/smv/assign/assign1.desc
+++ b/regression/smv/assign/assign1.desc
@@ -1,0 +1,8 @@
+KNOWNBUG
+assign1.smv
+
+^file .* line 6: variable `x' already assigned.*$
+^EXIT=1$
+^SIGNAL=0$
+--
+--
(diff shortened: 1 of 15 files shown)


regression/smv/assign/assign1.desc
```diff
@@ -0,0 +1,8 @@
+KNOWNBUG
+assign1.smv
+
+^file .* line 6: variable `x' already assigned.*$
+^EXIT=1$
+^SIGNAL=0$
+--
+--
```
tautschnig, line 8: It seems there is one test that now works while the others are all `KNOWNBUG`: do those need further work?

tautschnig (approved):

regression/smv/assign/assign1.desc
kroening: Yes, will need more work.

PR #991, opened 2025-02-17
SystemVerilog: floating-point casts

kroening:
Casts from `real`/`shortreal`/`realtime` to integers must use round-to-nearest-ties-to-away.
    
The standard doesn't specify how to round int-to-float or double-to-single. We use RNA for consistency.

### DIFF
--- a/src/verilog/verilog_lowering.cpp
+++ b/src/verilog/verilog_lowering.cpp
@@ -251,19 +251,44 @@ exprt verilog_lowering_cast(typecast_exprt expr)
   auto &src_type = expr.op().type();
   auto &dest_type = expr.type();
 
+  // float to int
+  if(
+    (src_type.id() == ID_verilog_real ||
+     src_type.id() == ID_verilog_shortreal ||
+     src_type.id() == ID_verilog_realtime || src_type.id() == ID_floatbv) &&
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_lowering.cpp
```diff
@@ -251,19 +251,44 @@ exprt verilog_lowering_cast(typecast_exprt expr)
...
+    // to single precision floating-point.  It does not say how
+    // to round integers to float.  We simply use RNA,
+    // given the explicit requirement to use RNA when converting
+    // float to integer.
     expr.type() = verilog_lowering(dest_type);
     auto new_cast = floatbv_typecast_exprt{
       expr.op(),
       ieee_floatt::rounding_mode_expr(
-        ieee_floatt::rounding_modet::ROUND_TO_EVEN),
+        ieee_floatt::rounding_modet::ROUND_TO_AWAY),
```
tautschnig, line 291: Wouldn't this rather call for an "unspecified" rounding mode here when aiming for a sound verification tool?

kroening: I don't believe that there is any FV tool that deals with floating-point types; this is an esoteric feature.

tautschnig (approved):

PR #995, opened 2025-02-18
ebmc: `--ignore-initial`

kroening:
This adds the `--igore-initial` command line option to ebmc, which allows reasoning about the transition relation only of a Verilog file.


tautschnig:
Needs conflict resolution.

tautschnig (approved):
I'd appreciate a test that uses this option. Perhaps a variation on the one that had a file renamed?

kroening:
> I'd appreciate a test that uses this option. Perhaps a variation on the one that had a file renamed?

Forgot the git add, now fixed.

PR #999, opened 2025-02-18
SMV: avoid some expression copies

kroening:
References are changed to rvalue references to avoid some copying when populating the parse tree.

### DIFF
--- a/src/smvlang/parser.y
+++ b/src/smvlang/parser.y
@@ -486,17 +486,14 @@ assignments: assignment
 
 assignment : assignment_head '(' assignment_var ')' BECOMES_Token formula ';'
            {
-             binary($$, $3, ID_equal, $6);
-
              if(stack_expr($1).id()==ID_smv_next)
              {
-               exprt &op=to_binary_expr(stack_expr($$)).op0();
-               unary_exprt tmp(ID_smv_next, std::move(op));
-               tmp.swap(op);
-               PARSER.module->add_assign_next(to_equal_expr(stack_expr($$)));
+               PARSER.module->add_assign_next(
+                 unary_exprt{ID_smv_next, std::move(stack_expr($3))},
+                 std::move(stack_expr($6)));
              }
              else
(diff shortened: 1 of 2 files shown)


src/smvlang/parser.y
```diff
@@ -486,17 +486,15 @@ assignments: assignment
...
-             binary($$, $3, ID_equal, $6);
-
              if(stack_expr($1).id()==ID_smv_next)
              {
-               exprt &op=to_binary_expr(stack_expr($$)).op0();
+               exprt &op=stack_expr($3);
                unary_exprt tmp(ID_smv_next, std::move(op));
                tmp.swap(op);
-               PARSER.module->add_assign_next(to_equal_expr(stack_expr($$)));
+               PARSER.module->add_assign_next(std::move(stack_expr($3)), std::move(stack_expr($6)));
```
tautschnig, line 494: How about:
```
unary_exprt tmp{ID_smv_next, stack_expr($3)};
PARSER.module->add_assign_next(std::move(tmp), std::move(stack_expr($6)));
```
even if just a human not having to track what really lives at `$3` at which point in this sequence.

src/smvlang/smv_parse_tree.h
```diff
@@ -118,100 +118,110 @@ class smv_parse_treet
...
+    void
+    add_item(itemt::item_typet item_type, exprt expr, source_locationt location)
     {
-      items.push_back(itemt());
+      items.push_back(itemt{});
       items.back().item_type=item_type;
-      items.back().expr=expr;
-      items.back().location=location;
+      items.back().expr = std::move(expr);
+      items.back().location = std::move(location);
```
tautschnig, line 127: Perhaps `itemt` should have a proper constructor, at which point it should be possible to use
`items.emplace_back(item_type, std::move(expr), std::move(location));`. (Which in turns makes me wonder whether we could then just get rid of `add_item` as a wrapper altogether.)

tautschnig (approved):

PR #1001, opened 2025-02-20
SystemVerilog: `--initial-zero`

kroening:
The SystemVerilog standard specifies that variables are to be zero initialized before simulation begins.  Some synthesis tools generate logic to achieve this, whereas others do not.

This adds the option `--initial-zero` to match the synthesis semantics when appropriate.

### DIFF
--- a/src/verilog/verilog_initializer.cpp
+++ b/src/verilog/verilog_initializer.cpp
@@ -1,0 +1,49 @@
+/*******************************************************************\
+
+Module: Verilog Initializer
+
+Author: Daniel Kroening, dkr@amazon.com
+
+\*******************************************************************/
+
+#include "verilog_initializer.h"
+
+#include <util/arith_tools.h>
+#include <util/std_expr.h>
+
+std::optional<exprt> verilog_default_initializer(const typet &type)
+{
+  if(type.id() == ID_signedbv || type.id() == ID_unsignedbv)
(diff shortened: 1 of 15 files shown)


src/verilog/verilog_initializer.cpp
```diff
@@ -0,0 +1,44 @@
...
+      auto member_value_opt = verilog_zero_initializer(component.type());
+      if(!member_value_opt.has_value())
+        return {};
+      member_values.push_back(*member_value_opt);
+    }
+    return struct_exprt{std::move(member_values), struct_type};
+  }
+  else
+    return {};
+}
```
tautschnig, line 44: It would seem safe to just use `zero_initializer` from CBMC's code base?

tautschnig: Ping? ^^

tautschnig: Do we actually need to make CBMC's `zero_initializer` easier to extend?

kroening: Yes, it's either that or a copy.

kroening:
> I'm always opposed to code duplication so I'd strongly favour an approach that will facilitate code reuse, but I understand that this may require more effort than is possible to put in right now.

We also need extensibility for the solvers, the SMT-LIB converter, and the simplifier.

PR #1010, opened 2025-02-25
SMV netlist: mark properties that were not translated

kroening:
When a property cannot be mapped to a netlist property, add a comment to the SMV file saying so.


src/trans-netlist/netlist.cpp
```diff
@@ -368,6 +368,12 @@ void netlistt::output_smv(std::ostream &out) const
       print_smv(out, std::get<GFpt>(property).p);
       out << '\n';
     }
+    else
+    {
+      out << "-- " << id << '\n';
+      out << "-- not translated\n";
+      out << '\n';
+    }
```
tautschnig, line 376: Wouldn't it be better to use `else if` here and have an `else UNREACHABLE;`?

tautschnig (approved):

PR #1017, opened 2025-03-04
CI: Ubuntu 20.04 -> 24.04

kroening:
Github will remove the Ubuntu 20.04 LTS runner on 2025-04-01.  This moves the existing jobs from 20.04 to 24.04.

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -7,8 +7,8 @@ on:
 
 jobs:
   # This job takes approximately 15 minutes
-  check-ubuntu-20_04-make-gcc:
-    runs-on: ubuntu-20.04
+  check-ubuntu-24_04-make-gcc:
+    runs-on: ubuntu-24.04
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request-checks.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           sudo apt-get update
-          sudo apt-get install --no-install-recommends -yq clang-10 clang++-10 gdb jq flex bison libxml2-utils cpanminus ccache z3
+          sudo apt-get install --no-install-recommends -yq clang-19 clang++-19 gdb jq flex bison libxml2-utils cpanminus ccache z3
```
tautschnig, line 81: Perhaps this should now just be `clang` and `clang++` instead of using a particular version of the compiler?

tautschnig (approved):

.github/workflows/pull-request-checks.yaml
kroening: That was my first attempt -- `apt-get install` errors.

PR #1036, opened 2025-03-26
Fix unused variables that only appear in `assert`

mgudemann:
This fixes instances in the IC3 implementation where variables are only used for `assert`, leading to errors when compiling with `g++` 13.3.0.
This also removes `assert(false)` in two instances in the IC3 implementation.

### DIFF
--- a/src/ic3/aux_types.hh
+++ b/src/ic3/aux_types.hh
@@ -6,8 +6,6 @@ Author: Eugene Goldberg, eu.goldberg@gmail.com
 
 ******************************************************/
 
-#pragma once
-
 #include <queue>
 #include <string>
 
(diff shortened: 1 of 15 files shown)


tautschnig (approved):
Still needs formatting fixes, and https://github.com/diffblue/hw-cbmc/pull/1037 should help so that the formatting guidance is provided in the GitHub action log.

src/ic3/aux_types.hh
```diff
@@ -6,8 +6,14 @@ Author: Eugene Goldberg, eu.goldberg@gmail.com
 
 ******************************************************/
 
+#pragma once
```
tautschnig, line 9: We don't usually use this in this code base. Why is it necessary here?

tautschnig: None of those ic3 headers appear to use a header guard, so omit altogether if it compiles, I'd say.

PR #1043, opened 2025-03-30
BDD engine: only look for atomic propositions in supported properties

kroening:
The BDD engine is changed to avoid searching unsupported properties for atomic propositions.

### DIFF
--- a/src/ebmc/bdd_engine.cpp
+++ b/src/ebmc/bdd_engine.cpp
@@ -65,6 +65,8 @@ protected:
   const namespacet ns;
   netlistt netlist;
 
+  static bool property_supported(const exprt &);
+
   // the Manager must appear before any BDDs
   // to do the cleanup in the right order
   mini_bdd_mgrt mgr;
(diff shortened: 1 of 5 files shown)


src/ebmc/bdd_engine.cpp
```diff
@@ -455,6 +504,12 @@ void bdd_enginet::check_property(propertyt &property)
   if(property.is_assumed())
     return;
 
+  if(!property_supported(property.normalized_expr))
+  {
+    property.failure("property not supported by BDD engine");
+    return;
+  }
+
```
tautschnig, line 512: Should the code that follows (the code that actually does the transform to CTL) be combined with the `property_supported` code? That is, should `property_supported` right away do the transform?

src/temporal-logic/ctl.cpp
```diff
@@ -0,0 +1,16 @@
...
+\*******************************************************************/
+
+#include "ctl.h"
+
+#include "temporal_logic.h"
+
+bool is_AGp(const exprt &expr)
+{
+  return expr.id() == ID_AG && !has_temporal_operator(to_AG_expr(expr).op());
+}
```
tautschnig, line 16: Surprisingly, this is different from `is_AGp` in `bdd_enginet::check_property`?!

kroening: Now clearer, avoiding labeling expressions as AGp that are just equivalent.

src/ebmc/bdd_engine.cpp
kroening: The duplication is also bugging me, but it seems even less desirable to make the creation of some new data structure a side-effect of `property_supported`.

tautschnig: How about making `property_supported` (though then perhaps by the name of `make_CTL_property` or the likes) return a `std::optional<exprt>` where an unsupported property would cause `{}` to be returned?

kroening: Coming with #1051

tautschnig (approved):
Approving with the assumption that #1051 will bring about further improvements.

PR #1046, opened 2025-03-31
SVA: `require_sva_property`

kroening:
This introduces a method to check operands of SVA expressions that are required to be an SVA property.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.h
+++ b/src/verilog/verilog_typecheck_expr.h
@@ -200,6 +200,7 @@ protected:
   }
 
   void require_sva_sequence(const exprt &) const;
+  void require_sva_property(exprt &);
 
   [[nodiscard]] exprt convert_sva_rec(exprt);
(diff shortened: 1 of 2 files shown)


src/verilog/verilog_typecheck_expr.h
```diff
@@ -200,6 +200,7 @@ class verilog_typecheck_exprt:public verilog_typecheck_baset
   }
 
   void require_sva_sequence(const exprt &) const;
+  void require_sva_property(exprt &);
```
tautschnig, line 203: It's a bit weird that only the latter has a side effect.

tautschnig (approved):

src/verilog/verilog_typecheck_expr.h
kroening: Coming as a separate PR.

kroening: #1047

PR #1051, opened 2025-04-04
SMV netlist: translate some SVA to LTL

kroening:
Parts of SVA map directly to LTL.  This allows extending the set of formulas converted to LTL when outputting an SMV model.

### DIFF
--- a/src/temporal-logic/temporal_logic.cpp
+++ b/src/temporal-logic/temporal_logic.cpp
@@ -210,3 +210,79 @@ std::optional<exprt> LTL_to_CTL(exprt expr)
   else
     return {};
 }
+
+std::optional<exprt> SVA_to_LTL(exprt expr)
+{
+  // Some SVA is directly mappable to LTL
+  if(expr.id() == ID_sva_always)
+  {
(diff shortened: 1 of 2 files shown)


src/temporal-logic/temporal_logic.cpp
```diff
@@ -152,3 +154,78 @@ bool is_SVA_always_s_eventually_p(const exprt &expr)
...
+    {
+      auto rec = SVA_to_LTL(op);
+      if(!rec.has_value())
+        return {};
+      op = rec.value();
+    }
+    return copy;
+  }
+  else if(!has_temporal_operator(expr))
+    return expr;
```
tautschnig, line 228: Wouldn't moving this case before the `else if(expr.id() == ID_and ...)` avoid unnecessary copying when it is a conjunction/disjunction/etc. without temporal operators?

tautschnig (approved):

src/temporal-logic/temporal_logic.cpp
kroening: Yes, will move up

PR #1061, opened 2025-04-10
simplify `false R ψ` to `G ψ`

kroening:
This simplification enables printing shorter counterexample traces.

### DIFF
--- a/src/temporal-logic/normalize_property.cpp
+++ b/src/temporal-logic/normalize_property.cpp
@@ -103,6 +103,14 @@ exprt normalize_property_rec(exprt expr)
     op = normalize_property_rec(op); // recursive call
 
   // post-traversal
+  if(expr.id() == ID_R)
+  {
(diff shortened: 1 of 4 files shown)


src/temporal-logic/normalize_property.cpp
```diff
@@ -103,6 +103,14 @@ exprt normalize_property_rec(exprt expr)
     op = normalize_property_rec(op); // recursive call
 
   // post-traversal
+  if(expr.id() == ID_R)
+  {
+    if(to_R_expr(expr).lhs().is_false())
+    {
+      // false R ψ ≡ G ψ
```
tautschnig, line 110: normalize_property.h documents the collection of normalisation steps applied, so I think this should be added to that documentation.

tautschnig (approved):

src/temporal-logic/normalize_property.cpp
kroening: added

PR #1063, opened 2025-04-12
SystemVerilog: `const`

kroening:
This adds a check that the LHS of assignments is not `const`.


src/hw_cbmc_irep_ids.h
```diff
@@ -94,6 +94,7 @@ IREP_ID_ONE(sva_implies)
 IREP_ID_ONE(sva_not)
 IREP_ID_ONE(sva_or)
 IREP_ID_ONE(module_instance)
+IREP_ID_TWO(C_const, #const)
```
tautschnig, line 97: Shouldn't we just use `C_constant` from CBMC? Over there, it's used for exactly the same purpose.

tautschnig: I agree with the mis-naming. I obviously don't intend to block this PR, but I'd prefer to see a medium-term cleanup of this across the two code bases.

PR #1066, opened 2025-04-15
SVA to LTL now converts SVA sequences

kroening:
This adds support for mapping some SVA sequences to equivalent LTL.

### DIFF
--- a/src/temporal-logic/temporal_logic.cpp
+++ b/src/temporal-logic/temporal_logic.cpp
@@ -225,6 +225,108 @@ static exprt n_Xes(mp_integer n, exprt op)
     return n_Xes(n - 1, X_exprt{std::move(op)});
 }
 
+// Returns a set of match conditions (given as LTL formula)
+struct ltl_sequence_matcht
+{
+  ltl_sequence_matcht(exprt __cond, mp_integer __length)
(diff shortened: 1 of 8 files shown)


tautschnig:
This one needs a rebase.

kroening:
Done

src/temporal-logic/temporal_logic.cpp
```diff
@@ -224,6 +224,85 @@ static exprt n_Xes(mp_integer n, exprt op)
...
+      for(auto &match : matches)
+      {
+        // delay as instructed
+        match.length += from_int;
+        match.cond = n_Xes(from_int, match.cond);
+      }
+      return matches;
+    }
+    else
+      return {};
```
tautschnig, line 296: Wouldn't the upper bound on `delay` just mean that after such delay we want the negation of `match.cond` to hold?

tautschnig (approved):

src/temporal-logic/temporal_logic.cpp
kroening: It only says that the match has to happen in that interval, no claim is made about what happens after that interval.

I'll add a `for` loop.

PR #1068, opened 2025-04-18
bugfix for SVA cover

kroening:
This changes the implicit 'always' used for SVA cover properties from LTL G to SVA `always`.  This avoids a mixture of LTL and SVA, which is not supported by the BMC backend.

### DIFF
--- a/src/temporal-logic/normalize_property.cpp
+++ b/src/temporal-logic/normalize_property.cpp
@@ -119,7 +119,7 @@ exprt normalize_property(exprt expr)
 {
   // top-level only
   if(expr.id() == ID_sva_cover)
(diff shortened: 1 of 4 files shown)


src/temporal-logic/normalize_property.cpp
```diff
@@ -119,7 +119,7 @@ exprt normalize_property(exprt expr)
 {
   // top-level only
   if(expr.id() == ID_sva_cover)
-    expr = G_exprt{not_exprt{to_sva_cover_expr(expr).op()}};
+    expr = sva_always_exprt{not_exprt{to_sva_cover_expr(expr).op()}};
```
tautschnig, line 122: I think we were missing the corresponding comment in `normalize_property.h` -- would you mind adding that one?

PR #1071, opened 2025-04-19
SVA: default sequence semantics

kroening:
SVA sequences can have weak or strong semantics.  1800-2017 16.12.2 Sequence property requires that sequences in properties that are not explicitly marked as strong or weak are interpreted as weak when used in `assert property` or `assume property`, and strong otherwise.

The Verilog type checker now emits corresponding expressions `sva_implict_weak` and `sva_implicit_strong` for uses of SVA sequences in properties.


kroening:
> Is there any way in which we'd actually notice a difference between these different semantics? My reading of the changes is that for now we just keep track of whatever was the input configuration, but ultimately still have just one semantics?

Yes, the use of it is in #1067

PR #1075, opened 2025-04-21
KNOWNBUG test for empty SVA sequences

kroening:


regression/verilog/SVA/empty_sequence1.desc
```diff
@@ -0,0 +1,8 @@
+KNOWNBUG
+empty_sequence1.sv
+--bound 5
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
```
tautschnig, line 8: Would you mind adding a note as to what the problem here is?

kroening: Done

tautschnig (approved):

PR #1078, opened 2025-04-21
aval/bval encoding for shifts

kroening:
This adds an aval/bval encoding for four-valued shift expressions to the Verilog frontend.


tautschnig:
Should the second commit change the "KNOWNBUG" status of the tests to "CORE"?

PR #1080, opened 2025-04-23
Introduce `sva_sequence_repetition_exprt`

kroening:
This introduces `sva_sequence_repetition_exprt` as a base class for the SVA repetition operators `[*...]`, `[=...]`, `[->...]`.

### DIFF
--- a/src/verilog/verilog_typecheck_sva.cpp
+++ b/src/verilog/verilog_typecheck_sva.cpp
@@ -270,26 +270,6 @@ exprt verilog_typecheck_exprt::convert_binary_sva(binary_exprt expr)
 
     return std::move(expr);
   }
-  else if(
-    expr.id() == ID_sva_sequence_non_consecutive_repetition ||
-    expr.id() == ID_sva_sequence_goto_repetition)
-  {
-    convert_sva(expr.lhs());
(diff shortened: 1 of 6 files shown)


src/verilog/verilog_typecheck_sva.cpp
```diff
@@ -342,22 +322,41 @@ exprt verilog_typecheck_exprt::convert_ternary_sva(ternary_exprt expr)
...
+    convert_expr(expr.op1());
+
+    mp_integer n = elaborate_constant_integer_expression(expr.op1());
+    if(n < 0)
+      throw errort().with_location(expr.op1().source_location())
+        << "number of repetitions must not be negative";
+
+    expr.op1() = from_integer(n, integer_typet{});
+
+    expr.type() = verilog_sva_sequence_typet{};
```
tautschnig, line 356: Is this even necessary anymore?

tautschnig (approved):
Requires conflict resolution now that #1082 has been merged.

src/verilog/verilog_typecheck_sva.cpp
kroening: Unfortunately yes, since the parser doesn't use the expression constructors.

PR #1085, opened 2025-04-24
convert `s_until` and `s_until_with` to LTL

kroening:
This adds conversion for the SVA operators `s_until` and `s_until_with` to LTL.


tautschnig (approved):
Now that #1089 has gone in, could we even have a test for this one?

kroening:
> Now that #1089 has gone in, could we even have a test for this one?

Yes, that was the plan

PR #1090, opened 2025-04-26
use `sva_sequence_repetition_exprt` for `[*]` and `[+]`

kroening:
The base class `sva_sequence_repetition_exprt` is now also used for the SVA consecutive sequence repetition operators `[*]` and `[+]`.


tautschnig:
Needs a rebase.

kroening:
> Needs a rebase.

done

src/trans-word-level/sequence.cpp
```diff
@@ -364,42 +364,52 @@ sequence_matchest instantiate_sequence(
...
+      // [*x:$]
+      auto from = numeric_cast_v<mp_integer>(repetition.from());
+      auto &op = repetition.op();
 
-    // Is x a sequence or a state predicate?
-    if(is_SVA_sequence_operator(op))
-      PRECONDITION(false); // no support
+      // Is op a sequence or a state predicate?
+      if(is_SVA_sequence_operator(op))
+        PRECONDITION(false); // no support
```
tautschnig, line 378: Is this the right way to report this problem, or should this be an exception (with a human-readable error message) instead? (I acknowledge that this is the same behaviour as before.)

tautschnig (approved):

PR #1096, opened 2025-05-02
BDD engine: transform SVA to LTL, then LTL to CTL

kroening:
The BDD engine now transforms SVA to LTL, when possible, which in turn is transformed to CTL, when possible.

This is strictly more general than the current direct mapping.


regression/verilog/SVA/sequence_repetition1.sv
```diff
@@ -1,13 +1,13 @@
 module main(input clk);
 
-  reg [31:0] x = 0;
+  reg [7:0] x = 0;
```
tautschnig, line 3: Why do the changes to property transforms entail this change? AFAIK this test was passing before.

kroening: They now do run with BDDs, and the BDDs simply don't scale to 32 bits. They can do 8.

tautschnig: But then how did this test pass before?

kroening: We did SAT-based BMC only. That can do 32 bits.

tautschnig (approved):

PR #1104, opened 2025-05-07
KNOWNBUG test for `elsif

kroening:


regression/verilog/preprocessor/elsif1.desc
```diff
@@ -0,0 +1,6 @@
+KNOWNBUG
+elsif1.v
+
+^EXIT=0$
+^SIGNAL=0$
+--
```
tautschnig, line 6: Can we have a line saying what the problem is?

tautschnig (approved):

PR #1114, opened 2025-05-18
SVA-to-LTL: sequence `and`/`or`

kroening:
This adds SVA sequence `and`/`or` to the SVA-to-LTL translator.


tautschnig:
Needs git conflict resolution.

kroening:
> Needs git conflict resolution.

Done

tautschnig (commented):
I am assuming the complexity of the ID_sva_and case is just unavoidable (or maybe those matches are always small?).

tautschnig (approved):

kroening:
> I am assuming the complexity of the ID_sva_and case is just unavoidable (or maybe those matches are always small?).

The real answer is to make an automaton, and merge that with the model. I don't see a better way when the goal is "readable, equivalent LTL".

PR #1121, opened 2025-05-21
Add a Dockerfile for building and running EBMC

Po-Chun-Chien:
This Dockerfile uses a two-stage build process similar to [CBMC's Dockerfile](https://github.com/diffblue/cbmc/blob/develop/Dockerfile):
the binary is compiled in the first stage and then copied into another container in the second stage.

The build step follows the instructions in `COMPILING.md`,
with the assumption that the submodule `lib/cbmc` has already been checked out locally.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM ubuntu:24.04 AS builder
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update
+RUN apt-get upgrade -y
+RUN apt-get install -y --no-install-recommends \
+    g++ \
+    gcc \
+    flex \
+    bison \
+    make \
+    curl \
+    patch
+
+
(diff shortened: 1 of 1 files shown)


kroening:
May I suggest installing `git` in the container -- this would enable #1121 to work.

Po-Chun-Chien:
> May I suggest installing `git` in the container -- this would enable #1121 to work.

Sure, done! I assume you meant #1118, not #1121.
I also tested the Docker file at that branch (`git-version`)
and the version string with git commit is correctly displayed.

```bash
$ podman run --rm localhost/ebmc-version --version
5.7 (ebmc-5.6-21-g840d47fa)
```

kroening:
Thank you -- all that's missing is squashing the two commits together, and this is good to merge.

kroening (approved):

PR #1128, opened 2025-05-29
SVA: use `from`/`to` instead of `lower`/`upper`

kroening:
This changes the names of methods and variables for the two end-points of SVA predicate ranges to read `from` for the first endpoint, and `to` for the second, instead of `lower` and `upper`.

This avoids a collision with the term `lower` for "lowering".


tautschnig (approved):
Requires conflict resolution.

PR #1130, opened 2025-05-30
SVA-to-Buechi: sequence repetition operators

kroening:
This adds conversion for `[*...]`, `[=...]`, `[->...]`, `[+]` to the SVA-to-Buechi translation.


regression/ebmc-spot/sva-buechi/sequence_repetition6.desc
```diff
@@ -1,8 +0,0 @@
-CORE
```
tautschnig, line 1: Why was this one deleted?

kroening: The behavior it tests is the same as regression/verilog/SVA/sequence_repetition6.desc

PR #1131, opened 2025-05-30
SVA-to-Buechi: `s_always`

kroening:
This adds conversion for SVA `s_always` to the SVA-to-Buechi translation.


tautschnig (approved):
A test would be appreciated (and care is required that this one will also need the `lower` -> `from` change.

kroening:
> A test would be appreciated (and care is required that this one will also need the `lower` -> `from` change.

Test added.

PR #1136, opened 2025-06-02
mark a test as broken-smt-backend

kroening:


tautschnig (approved):

tautschnig:
Is this a matter of the Z3 version being used?

kroening:
> Is this a matter of the Z3 version being used?

That was my first suspicion, but I haven't been able to reproduce this locally yet. Locally, the test simply passes, with a wide range of z3 versions.

PR #1137, opened 2025-06-03
Verilog: mark unsized literals

kroening:
This marks unsized literals (e.g., `1`, `'b0101`, `'z`) as such, to enable rules that prohibit their use.


src/verilog/convert_literals.cpp
```diff
@@ -118,6 +118,20 @@ constant_exprt convert_real_literal(const irep_idt &value)
...
+static constant_exprt unsized(constant_exprt expr)
+{
+  expr.set(ID_C_verilog_unsized, true);
+  return expr;
+}
+
+static constant_exprt cond_unsized(constant_exprt expr, bool is_unsized)
+{
+  if(is_unsized)
+    return unsized(expr);
```
tautschnig:
```suggestion
    return unsized(std::move(expr));
```

PR #1139, opened 2025-06-04
word-level BMC: empty sequences

kroening:
SVA empty matches do not evaluate to true when used as property.

### DIFF
--- a/src/trans-word-level/sequence.h
+++ b/src/trans-word-level/sequence.h
@@ -19,12 +19,30 @@ class sequence_matcht
 {
 public:
   sequence_matcht(mp_integer __end_time, exprt __condition)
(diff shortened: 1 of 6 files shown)


src/trans-word-level/sequence.h
```diff
@@ -19,12 +19,27 @@ class sequence_matcht
...
+      condition(std::move(__condition))
   {
   }
 
+  bool empty_match() const
+  {
+    return _is_empty_match;
+  }
+
+  bool _is_empty_match;
```
tautschnig, line 34: This should be marked `protected` or `private`. (The same is true of the fields below, but those would need to get accessors.)

tautschnig (approved):

PR #1145, opened 2025-06-06
Test for SVA empty match and followed-by operator

kroening:
This adds a test that checks the behavior of an SVA empty match used as the LHS of the two followed-by operators.

### DIFF
--- a/regression/ebmc-spot/sva-buechi/followed-by5.bdd.desc
+++ b/regression/ebmc-spot/sva-buechi/followed-by5.bdd.desc
@@ -1,0 +1,11 @@
+KNOWNBUG
+followed-by5.sv
+--buechi --bdd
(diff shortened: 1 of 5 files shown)


regression/ebmc-spot/sva-buechi/followed-by5.bdd.desc
```diff
@@ -0,0 +1,11 @@
+KNOWNBUG
+followed-by5.sv
```
tautschnig: Not in this folder

regression/ebmc-spot/sva-buechi/followed-by5.bmc.desc
```diff
@@ -0,0 +1,11 @@
+KNOWNBUG
+followed-by5.sv
```
tautschnig: Not in this folder

PR #1149, opened 2025-06-09
Verilog: `$isunknown`

kroening:
This adds a lowering and constant-folding for `$isunknown`.


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -488,6 +489,31 @@ exprt verilog_typecheck_exprt::convert_expr_function_call(
...
+
+ Purpose:
+
+\*******************************************************************/
+
+constant_exprt verilog_typecheck_exprt::isunknown(const constant_exprt &expr)
+{
+  // constant folding for $isunknown
+  auto bval = ::bval(expr);
+  auto bval_simplified = simplify_expr(bval, ns);
```
tautschnig: Should this possibly use `verilog_simplifier`?

tautschnig (approved):

src/verilog/verilog_typecheck_expr.cpp
kroening: Done

PR #1154, opened 2025-06-12
introduce SMV enumeration type

kroening:
SMV enumerations behave differently from `enumeration_typet`; hence, this introduces `smv_enumeration_typet`, which is subsequently lowered to `enumeration_typet`.


src/smvlang/smv_types.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+bool smv_enumeration_typet::is_subset_of(
+  const smv_enumeration_typet &other) const
+{
+  // This is quadratic.
+  for(auto &element1 : elements())
+  {
+    bool found = false;
+    for(auto &element2 : other.elements())
```
tautschnig: Given the realistic size of enums the quadratic algorithm might not matter too much, but repeatedly invoking `other.elements()` seems avoidable. This should be `other.find(ID_elements).get_sub()` instead.

kroening: Now replaced by a set, which makes this n log n.

src/smvlang/smv_types.cpp
```diff
@@ -0,0 +1,82 @@
...
+bool smv_enumeration_typet::is_subset_of(
+  const smv_enumeration_typet &other) const
+{
+  std::set<irep_idt> other_elements;
+  auto &other_elements_sub = other.find(ID_elements).get_sub();
+
+  for(auto &element : other_elements_sub)
+    other_elements.insert(element.id());
+
+  for(auto &element : elements())
```
tautschnig: Thank you for removing the quadratic approach. But I don't think we should use `elements()` here either as `for(auto &e : find(ID_elements).get_sub())` would avoid constructing the vector.

tautschnig (approved):

src/smvlang/smv_types.cpp
kroening: done

PR #1160, opened 2025-06-17
add (optional) lhs to SVA cycle delay operator

kroening:
The SVA cycle delay operator `##[...]` can have an optional left-hand side operand.  This adds an operand to the expression, which is nil when not used.


kroening:
The tests come for free with a separate PR, which will replace `sva_sequence_concatenation_exprt` by this new form of the cycle delay expression.

PR #1169, opened 2025-06-23
SVA: `sequence()` methods for first_match and throughout

kroening:
This adds convenience accessors for the sequence-typed operand of the SVA `first_match` and `throughout` expression classes.


kroening:
> Shouldn't those methods also be used?

Done

PR #1175, opened 2025-06-29
SMV: set type

kroening:
This fixes typechecking for set literals, assignments from set literals to variables of the element type, and typechecking for the `in` set membership operator.


src/smvlang/expr2smv.cpp
```diff
@@ -656,6 +656,9 @@ expr2smvt::resultt expr2smvt::convert_rec(const exprt &src)
   else if(src.id() == ID_mod)
     return convert_binary(to_mod_expr(src), src.id_string(), precedencet::MULT);
 
+  else if(src.id() == ID_smv_set)
+    return convert_nondet_choice(src);
```
tautschnig: This looks strange. Is this just a re-use of a function that wasn't meant to be used that way?

src/smvlang/expr2smv.cpp
```diff
@@ -791,9 +794,6 @@ expr2smvt::resultt expr2smvt::convert_rec(const exprt &src)
   else if(src.id()=="smv_nondet_choice")
     return convert_nondet_choice(src);
 
-  else if(src.id() == ID_constraint_select_one)
-    return convert_nondet_choice(src);
```
tautschnig, line 795: Why is this now gone?

src/smvlang/expr2smv.cpp
kroening: Renamed the method

PR #1188, opened 2025-07-07
Buechi: mark unconverted properties as unsupported

kroening:
The Buechi flow now marks properties that cannot be converted into a Buechi acceptance condition as unsupported.


tautschnig:
Does this also address any of the exisiting tests that were marked `KNOWNBUG` for yielding a wrong result?

kroening:
> Does this also address any of the exisiting tests that were marked `KNOWNBUG` for yielding a wrong result?

Unfortunately not, these seem to be different issues.

PR #1190, opened 2025-07-07
Unit tests for `trivial_sva(...)`

kroening:
This adds unit tests for the `trivial_sva(expr)` simplification.

### DIFF
--- a/unit/temporal-logic/trivial_sva.cpp
+++ b/unit/temporal-logic/trivial_sva.cpp
@@ -1,0 +1,60 @@
+/*******************************************************************\
+
+Module: Trivial SVA
+
+Author: Daniel Kroening, dkr@amazon.com
+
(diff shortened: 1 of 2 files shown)


unit/temporal-logic/trivial_sva.cpp
```diff
@@ -0,0 +1,64 @@
...
+Module: Trivial SVA
+
+Author: Daniel Kroening, dkr@amazon.com
+
+\*******************************************************************/
+
+//#include <util/arith_tools.h>
+//#include <util/mathematical_types.h>
+
+//#include <temporal-logic/ltl.h>
```
tautschnig: I'd suggest to remove those - they are easy enough to add back in when actually required.

tautschnig (approved):

PR #1195, opened 2025-07-11
SMV: complex identifiers are now parsed as expressions

kroening:
NuSMV allows a subset of the expression syntax to be used as variable identifiers.  This changes the SMV grammar to use expressions to store the result of the `complex_identifier` rule.

### DIFF
--- a/src/smvlang/parser.y
+++ b/src/smvlang/parser.y
@@ -130,6 +130,41 @@ static void j_binary(YYSTYPE & dest, YYSTYPE & op1, const irep_idt &id, YYSTYPE 
 
 /*******************************************************************\
 
+Function: merge_complex_identifier
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+irep_idt merge_complex_identifier(const exprt &expr)
+{
+  if(expr.id() == ID_symbol)
+    return to_symbol_expr(expr).get_identifier();
+  else if(expr.id() == ID_member)
+  {
+    auto &member_expr = to_member_expr(expr);
+    return id2string(merge_complex_identifier(member_expr.compound())) + '.' + id2string(member_expr.get_component_name());
+  }
+  else if(expr.id() == ID_index)
+  {
+    auto &index_expr = to_index_expr(expr);
+    auto &index = index_expr.index();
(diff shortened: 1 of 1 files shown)


src/smvlang/parser.y
```diff
@@ -130,6 +130,41 @@ static void j_binary(YYSTYPE & dest, YYSTYPE & op1, const irep_idt &id, YYSTYPE
...
+\*******************************************************************/
+
+irep_idt merge_complex_identifier(const exprt &expr)
+{
+  if(expr.id() == ID_symbol)
+    return to_symbol_expr(expr).get_identifier();
+  else if(expr.id() == ID_member)
+  {
+    auto &member_expr = to_member_expr(expr);
+    return id2string(merge_complex_identifier(member_expr.compound())) + '.' + id2string(member_expr.get_component_name());
```
tautschnig, line 150: In CBMC's field sensitivity encoding we chose `..` as separator to avoid confusing humans as to whether they see an identifier or an expression. Would that also be a possible concern (and solution) here?

src/smvlang/parser.y
```diff
@@ -130,6 +130,41 @@ static void j_binary(YYSTYPE & dest, YYSTYPE & op1, const irep_idt &id, YYSTYPE
...
+  }
+  else if(expr.id() == ID_index)
+  {
+    auto &index_expr = to_index_expr(expr);
+    auto &index = index_expr.index();
+    PRECONDITION(index.is_constant());
+    auto index_string = id2string(to_constant_expr(index).get_value());
+    return id2string(merge_complex_identifier(index_expr.array())) + '.' + index_string;
```
tautschnig, line 158: In keeping with the above comment: would `"[[" + index_string + "]]"` be a more human-friendly choice?

tautschnig (approved):

src/smvlang/parser.y
kroening: I did worry about this, but SMV doesn't have a `struct` type.

src/smvlang/parser.y
kroening: SMV does have an array type, and confusingly, uses the same syntax for those identifiers and arrays. Using say `[[` would make the difference clear, but then be incompatible with NuSMV.

PR #1198, opened 2025-07-12
BMC: exit early if there is no supported property

kroening:
BMC now exits early if there is no supported property, to prevent wasting time on generating the SAT instance.


regression/ebmc/engine-heuristic/unsupported1.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+unsupported1.smv
+
+^\[.*\] EX TRUE: UNKNOWN$
```
tautschnig, line 4: I suppose this also was the output before - shouldn't this .desc file specifically check for "No supported property" in the output?

kroening: Yes; will be a separate PR, since it's a change to the engine heuristic, not to the BMC engine.

PR #1202, opened 2025-07-13
BMC: reconsider lasso-shaped encodings

kroening:
This replaces the lasso-shaped encodings in the word-level BMC property encoding.  The lasso and straigth-line encodings are now built as separate obligations.  In the lasso encoding, all operators follow the same lasso shape.  This avoids spurious traces where the eventually operator follows the lasso but a safety part of the property does not.

### DIFF
--- a/src/smvlang/smv_typecheck.cpp
+++ b/src/smvlang/smv_typecheck.cpp
@@ -1940,9 +1940,13 @@ void smv_typecheckt::create_var_symbols(
       else
         symbol.pretty_name = strip_smv_prefix(symbol.name);
 
-      symbol.value = nil_exprt{};
-      symbol.is_input = true;
+      if(symbol.type.id() == "submodule")
+        symbol.is_input = false;
+      else
+        symbol.is_input = true;
+
       symbol.is_state_var = false;
+      symbol.value = nil_exprt{};
       symbol.location = item.expr.source_location();
 
       symbol_table.insert(std::move(symbol));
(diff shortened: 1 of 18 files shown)


src/smvlang/smv_typecheck.cpp
```diff
@@ -1947,6 +1947,17 @@ void smv_typecheckt::convert(smv_parse_treet::mc_varst &vars)
...
+    if(var.type.id() == "submodule")
+    {
+      symbol.is_input = false;
+      symbol.is_state_var = false;
+    }
+    else
+    {
+      symbol.is_input = true;
+      symbol.is_state_var = false;
+    }
```
tautschnig: This change requires git conflict resolution, but also I am confused by the change itself: `is_state_var` is `false` in both branches _and_ was set to `false` before already.

kroening: Now cleaned up

src/smvlang/smv_typecheck.cpp
```diff
@@ -1940,9 +1940,13 @@ void smv_typecheckt::create_var_symbols(
...
 
-      symbol.value = nil_exprt{};
-      symbol.is_input = true;
+      if(symbol.type.id() == "submodule")
+        symbol.is_input = false;
+      else
+        symbol.is_input = true;
+
       symbol.is_state_var = false;
+      symbol.value = nil_exprt{};
```
tautschnig, line 1949: This line and the one above aren't actually necessary, the `symbolt` constructor will have set these already. (Though lines 1973 and 1975 should then also be removed.)

tautschnig (approved):

src/smvlang/smv_typecheck.cpp
kroening: I think it helps with clarity. It's quite a bit of mental burden to remember what values are set by the various constructors.

PR #1213, opened 2025-07-26
SMV: use module items for `VAR` and `DEFINE` for creating symbols

kroening:
The SMV typechecker now uses the module items for creating the symbols for `VAR` and `DEFINE`.

### DIFF
--- a/src/smvlang/smv_typecheck.cpp
+++ b/src/smvlang/smv_typecheck.cpp
@@ -58,7 +58,7 @@ public:
   } modet;
 
   void convert(smv_parse_treet::modulet &);
-  void convert(smv_parse_treet::mc_varst &);
+  void create_var_symbols(const smv_parse_treet::modulet::item_listt &);
 
   void collect_define(const equal_exprt &);
   void convert_defines(exprt::operandst &invar);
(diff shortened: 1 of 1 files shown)


src/smvlang/smv_typecheck.cpp
```diff
@@ -1911,43 +1904,56 @@ Function: smv_typecheckt::convert_vars
...
+    {
+      typet type = item.expr.type();
 
-    // check the type, if any
-    if(type.is_not_nil())
+      // check the type
       check_type(type);
 
-    symbol.base_name = var_it.first;
+      symbol.base_name = to_symbol_expr(item.expr).get_identifier();
```
tautschnig: This re-use of the same symbolt storage makes me a bit uncomfortable. How about creating a fresh `symbolt` (then using the most suitable constructor) in here and then using `symbol_table.insert(std::move(symbol));` instead of `symbol_table.add`?

tautschnig (approved):

src/smvlang/smv_typecheck.cpp
kroening: ok, done

PR #1222, opened 2025-08-04
Add CI job that runs KNOWNBUG tests

tautschnig:
This is to make sure that we do not spuriously keep tests marked as "KNOWNBUG" when a source change fixes the problem.


kroening:
Could use the binary that is produced by the `check-ubuntu-24_04-make-clang` job?

tautschnig:
> Could use the binary that is produced by the `check-ubuntu-24_04-make-clang` job?

Indeed, this is now done!

PR #1223, opened 2025-08-04
LTL/SVA->Buechi: reject error traces early

kroening:
The Buechi automaton accepts counterexamples to the original property.  The Buechi acceptance condition requires a lasso where the loop contains an accepting state.  A proof of that requires satisfying the looping condition, which may require a very long counterexample.

This adds a shortcut using a sufficient condition for a counterexample: we can reject the property when we reach a state in which the Buechi automaton transitions into an accepting state that has an unconditional self-loop on it.  Such traces can always be extended to a lasso as above.

### DIFF
--- a/regression/ebmc-spot/sva-buechi/cover1.desc
+++ b/regression/ebmc-spot/sva-buechi/cover1.desc
@@ -1,6 +1,6 @@
 CORE
-cover1.sv
---bound 10 --numbered-trace
+../../verilog/SVA/cover1.sv
+--buechi --bound 10 --numbered-trace
 ^\[main\.p0\] cover main\.counter == 1: PROVED$
 ^Trace with 2 states:$
 ^main\.counter@0 = 0$
--- a/regression/ebmc-spot/sva-buechi/if1.bmc.desc
+++ b/regression/ebmc-spot/sva-buechi/if1.bmc.desc
@@ -1,6 +1,6 @@
 CORE
-if1.sv
---bound 2
+../../verilog/SVA/if1.sv
+--buechi --bound 2
 ^\[main\.p0\] always \(if\(main\.counter == 0\) nexttime main\.counter == 1\): PROVED up to bound 2$
 ^\[main\.p1\] always \(if\(main\.counter == 0\) nexttime main\.counter == 1 else nexttime main\.counter == 3\): REFUTED$
 ^EXIT=10$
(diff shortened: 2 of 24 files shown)


regression/ebmc-spot/sva-buechi/cover1.desc
```diff
@@ -0,0 +1,14 @@
...
+^Trace with 2 states:$
+^main\.counter@0 = 0$
+^main\.counter@1 = 1$
+^\[main\.p1\] cover main\.counter == 100: REFUTED up to bound 10$
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+Back-end support is missing.
```
tautschnig: What does this comment mean when the lines above suggest that the properties are proved and refuted, respectively?

regression/ebmc-spot/sva-buechi/if1.bmc.desc
```diff
@@ -0,0 +1,11 @@
...
+../../verilog/SVA/if1.sv
+--buechi --bound 2
+^\[main\.p0\] always \(if\(main\.counter == 0\) nexttime main\.counter == 1\): PROVED up to bound 2$
+^\[main\.p1\] always \(if\(main\.counter == 0\) nexttime main\.counter == 1 else nexttime main\.counter == 3\): REFUTED$
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+SMT-back end doesn't do cast from bool to bool.
```
tautschnig: How does this become apparent when this test isn't excluded from the tests run with the SMT back-end?

src/temporal-logic/ltl_to_buechi.cpp
```diff
@@ -155,6 +174,37 @@ ltl_to_buechi(const exprt &property, message_handlert &message_handler)
     message.debug() << "Buechi liveness signal: " << format(liveness_signal)
                     << messaget::eom;
 
+    // construct the error signal -- true when the next automaton state
+    // is nonaccepting with an unconditional self-loop.
```
tautschnig, line 178: Should this say "accepting" instead of "nonaccepting"?

regression/ebmc-spot/sva-buechi/cover1.desc
kroening: It's a leftover; removed

src/temporal-logic/ltl_to_buechi.cpp
kroening: Indeed; fixed

regression/ebmc-spot/sva-buechi/if1.bmc.desc
kroening: Another leftover; removed

PR #1225, opened 2025-08-06
EBMC: tautology checker

kroening:
This adds a property tautology check ask the first step to the engine selection.

While it should be rare that this pass proves a property, the fact that it does may suggest that the property is unintentionally vacuous.


src/ebmc/tautology_check.cpp
```diff
@@ -0,0 +1,81 @@
...
+      return true;
+
+    case decision_proceduret::resultt::D_SATISFIABLE:
+    case decision_proceduret::resultt::D_ERROR:
+    default:
+      return false;
+    }
+  }
+  else
+    return false;
```
tautschnig: Should this be `UNREACHABLE`?

tautschnig (approved):

src/ebmc/tautology_check.cpp
kroening: No, that's the case where there is a temporal operator, just not one of the ones handled above. I'll restructure the branches.

PR #1227, opened 2025-08-07
clean up HOA accepting states

kroening:
The HOAs generated by Spot can now be transformed into a graph.
    
This graph is used to remove accepting states that are not part of a cycle.

### DIFF
--- a/src/temporal-logic/hoa.cpp
+++ b/src/temporal-logic/hoa.cpp
@@ -543,3 +543,40 @@ void hoat::output(std::ostream &out) const
 
   out << "--END--\n";
 }
+
+grapht<graph_nodet<hoat::graph_edget>> hoat::graph() const
+{
+  grapht<graph_nodet<hoat::graph_edget>> graph;
+
+  graph.resize(numeric_cast_v<std::size_t>(max_state_number() + 1));
+
+  for(auto &state : body)
+  {
+    for(auto &edge : state.second)
+      for(auto &dest : edge.dest_states)
+      {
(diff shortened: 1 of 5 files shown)


src/temporal-logic/hoa.cpp
```diff
@@ -543,3 +543,22 @@ void hoat::output(std::ostream &out) const
...
+    for(auto &edge : state.second)
+      for(auto &dest : edge.dest_states)
+      {
+        graph.add_edge(state.first.number, dest);
+        graph.edge(state.first.number, dest).label = edge.label;
+      }
+  }
+
+  return graph;
+}
```
tautschnig, line 564: Based on out-of-band discussions we could now define
````c++
void hoat::buechi_acceptance_cleanup()
{
  using hoa_grapht = grapht<graph_nodet<hoat::graph_edget>>;
  auto as_graph = graph();
  for(auto &state : body)
  {
    if(state.is_accepting())
    {
      hoa_grapht::patht loop_path;
      as_graph.shortest_loop(state.first.number, loop_path);
      // when this state is not part of any cycle, then we cannot have Buechi acceptance via this state
      if(loop_path.empty())
        state.acc_sig.clear();
    }
  }
}
````

kroening: Added

tautschnig (approved):
Do we have any (existing) tests where we could demonstrate this works as intended?

kroening:
> Do we have any (existing) tests where we could demonstrate this works as intended?

Not easy to surface; I've added a unit test instead.

PR #1238, opened 2025-08-18
SystemVerilog: `++`/`--` for generate `for`

kroening:
SystemVerilog generate `for` allows both pre- and post-increment and -decrement.

### DIFF
--- a/src/verilog/verilog_generate.cpp
+++ b/src/verilog/verilog_generate.cpp
@@ -6,9 +6,10 @@ Author: Daniel Kroening, kroening@kroening.com
 
 \*******************************************************************/
 
-#include "verilog_typecheck.h"
+#include <util/arith_tools.h>
 
 #include "verilog_expr.h"
+#include "verilog_typecheck.h"
 
 /*******************************************************************\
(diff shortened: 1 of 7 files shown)


src/verilog/verilog_generate.cpp
```diff
@@ -268,7 +269,32 @@ void verilog_typecheckt::elaborate_generate_for(
...
+        auto new_value = binary_exprt{
+          op, statement == ID_preincrement ? ID_plus : ID_minus, one, typet{}};
+        auto assignment = verilog_generate_assignt{op, new_value};
+
+        elaborate_generate_assign(assignment, dest);
+      }
+      else
+      {
+        throw errort().with_location(for_statement.source_location())
+          << "unexpected genvar_iteration item";
```
tautschnig: Shouldn't this be some sort of `PRECONDITION` or `DATA_INVARIANT` instead? I don't think a user should be able to trigger this.

kroening: Replaced by `DATA_INVARIANT`

PR #1240, opened 2025-08-26
SystemVerilog `.*`

kroening:
SystemVerilog 1800-2005 has introduced the `.*` operator, used for

a) implicit port connections, and

b) patterns.

This adds support to the scanner, parser, and type checker.


CHANGELOG
```diff
@@ -4,6 +4,7 @@
 * Verilog: fix for typed parameter ports
 * SystemVerilog: fix for type parameters
 * SystemVerilog: type parameter ports
+* SystemVerilog: support 1800-2005 .* wildcard port connections
```
tautschnig: Is that really true? It's being parsed, but it's not really "supported" (or at least not by my definition of that term).

kroening: Yes, not there yet, fixed

PR #1247, opened 2025-08-28
`$past` now supports array and struct types

kroening:
`$past` returns a "zero value" when asked for a value prior to the initial frame.  This adds the zero value for array and struct types.

### DIFF
--- a/src/verilog/verilog_expr.cpp
+++ b/src/verilog/verilog_expr.cpp
@@ -15,6 +15,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <util/mathematical_types.h>
 #include <util/prefix.h>
 
+#include "verilog_initializer.h"
 #include "verilog_typecheck_base.h"
 #include "verilog_types.h"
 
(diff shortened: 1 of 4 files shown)


src/verilog/verilog_expr.cpp
```diff
@@ -533,11 +533,16 @@ exprt verilog_inside_exprt::lower() const
...
+    auto &array_type = to_array_type(type);
+    auto zero_rec = default_value(array_type.element_type());
+    return array_of_exprt{zero_rec, array_type};
+  }
   else
-    return std::move(zero);
+  {
+    return from_integer(0, type);
+  }
 }
```
tautschnig: Shouldn't this use CBMC's `zero_initializer` or else the one for Verilog being proposed in #1001?

kroening: Yes, the standard seems to guarantee that the value prior to the initial state is the default initial value, will consolidate.

kroening: All done

PR #1248, opened 2025-08-28
Verilog: reject use of SVA sequences and properties as Boolean expression

kroening:
SVA named sequences and named properties cannot be used as Booleans.  This adds a check.


tautschnig (commented):
What about regression/verilog/property/named_property5?

PR #1264, opened 2025-09-15
k-induction: precondition that no `$past` is present

kroening:
k-induction requires instrumentation of `$past`.  This adds a precondition that $past has been removed.


kroening:
The fix does come with this PR, and there is one test that catches this:
regression/ebmc/k-induction/past1.sv

PR #1271, opened 2025-09-24
Verilog: `ref` port direction

kroening:
SystemVerilog adds the `ref` port direction, usable for passing a reference to a variable to a module.

### DIFF
--- a/regression/verilog/modules/ref1.desc
+++ b/regression/verilog/modules/ref1.desc
@@ -1,0 +1,7 @@
+CORE
+ref1.sv
+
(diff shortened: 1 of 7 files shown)


regression/verilog/modules/ref1.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
```
tautschnig: Why is this still `KNOWNBUG`?

kroening: Fixed

PR #1275, opened 2025-09-25
Verilog: strengthen typing when using module type

kroening:
This adds the `ports()` member functions to the `module_typet` class, avoiding direct access to the components of the irep.

### DIFF
--- a/src/verilog/verilog_interfaces.cpp
+++ b/src/verilog/verilog_interfaces.cpp
@@ -50,8 +50,9 @@ Function: verilog_typecheckt::check_module_ports
 void verilog_typecheckt::check_module_ports(
   const verilog_module_sourcet::port_listt &module_ports)
 {
-  auto &ports = module_symbol.type.add(ID_ports).get_sub();
-  ports.resize(module_ports.size());
+  auto &ports = to_module_type(module_symbol.type).ports();
+  ports.clear();
+  ports.reserve(module_ports.size());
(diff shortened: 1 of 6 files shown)


src/verilog/verilog_interfaces.cpp
```diff
@@ -99,11 +99,11 @@ void verilog_typecheckt::check_module_ports(
...
     ports[nr].id(ID_symbol);
-    ports[nr].set(ID_identifier, identifier);
+    ports[nr].identifier(identifier);
     ports[nr].set(ID_C_source_location, declarator.source_location());
-    ports[nr].set(ID_type, port_symbol->type);
-    ports[nr].set(ID_input, port_symbol->is_input);
-    ports[nr].set(ID_output, port_symbol->is_output);
+    ports[nr].type() = port_symbol->type;
+    ports[nr].input(port_symbol->is_input);
+    ports[nr].output(port_symbol->is_output);
```
tautschnig: It would be nice to have a constructor to make sure necessary members are not forgotten upon initialisation.

kroening: I appreciate the goal of not forgetting to initialize a member, and that a constructor can enforce that.

On the downside, the code is less readable; e.g., you need to know the location of the two Booleans input/output in the list of parameters of the constructor. I am missing some syntax such as `constructor(.output = true, .input = false, ...)`

PR #1277, opened 2025-09-26
Verilog: Move shift tests

kroening:
This moves tests out of asic-world-operators/shift.sv to match the website.

### DIFF
--- a/regression/verilog/expressions/shl4.desc
+++ b/regression/verilog/expressions/shl4.desc
@@ -1,0 +1,9 @@
+KNOWNBUG
+shl4.sv
+--bound 0
+^EXIT=0$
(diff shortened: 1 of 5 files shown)


regression/verilog/expressions/shl4.desc
```diff
@@ -0,0 +1,9 @@
+KNOWNBUG
```
tautschnig, line 1: That's a bit scary that we are giving the wrong answer for such a basic test?

tautschnig (approved):

regression/verilog/expressions/shl4.desc
kroening: Yes; it's scary semantics.

PR #1290, opened 2025-10-01
Verilog: cleanout unused Boolean operators from type checker

kroening:
The Boolean `ID_xor`, `ID_xnor`, `ID_nand`, `ID_nor` are only ever generated as output by the Verilog type checker, and should never be inputs.

This removes the case that handles them as inputs.


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -3524,6 +3522,13 @@ exprt verilog_typecheck_exprt::convert_binary_expr(binary_exprt expr)
 
     return std::move(expr);
   }
+  else if(
+    expr.id() == ID_xor || expr.id() == ID_xnor || expr.id() == ID_nand ||
+    expr.id() == ID_nor)
+  {
+    // should not occur -- only generated by the typechecker
+    PRECONDITION(false);
+  }
```
tautschnig, line 3531: The `n_input_gatetype` in the parser has all these - should the parser be changed as well, or am I misreading this?

kroening: This is just the gate type, and doesn't turn into an expression before synthesis.

tautschnig (approved):

PR #1292, opened 2025-10-01
Verilog: cleanup downwards type propagation

kroening:
This cleans up `verilog_typecheck_exprt::downwards_type_progatation` to handle all the cases in 1800-2017 11.8.2.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -222,6 +222,39 @@ void verilog_typecheck_exprt::propagate_type(
 
 /*******************************************************************\
 
+Function: zero_extend
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+static exprt zero_extend(const exprt &expr, const typet &type)
+{
+  auto old_width = expr.type().id() == ID_bool ? 1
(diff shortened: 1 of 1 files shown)


tautschnig:
Should the failing regression test cause the test to be changed or is this a bug being introduced?

kroening:
I'll do the semantic change as a separate PR.

kroening:
Restored previous semantics, which prevents the test failure

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -222,6 +222,39 @@ void verilog_typecheck_exprt::propagate_type(
...
+  if(type.id() == ID_unsignedbv)
+    tmp_type = unsignedbv_typet{old_width};
+  else if(type.id() == ID_verilog_unsignedbv)
+    tmp_type = verilog_unsignedbv_typet{old_width};
+  else
+    PRECONDITION(false);
+
+  return typecast_exprt::conditional_cast(
+    typecast_exprt::conditional_cast(expr, tmp_type), type);
+}
```
tautschnig, line 254: Wouldn't use of `zero_extend_exprt` make this unnecessary?

kroening: Yes, please -- this will require more simplifier support, will come separately.

PR #1296, opened 2025-10-03
Generate VCD file for each cover statement

### LINKED ISSUE #1287: Generate VCD for each trace of cover property
EBMC can generate traces for each cover property using `--trace` switch, and each individual trace gets printed. Only 1 VCD file is getting generated with the `--vcd` switch, and I am not able to view each trace in a waveform viewer. 

It would be great if each trace gets generated as a separate VCD file, like how SBY in Yosys project currently does (https://github.com/YosysHQ/sby).

ShashankVM:
Fixes #1287


kroening:
This regresses on the current behavior, where you can get the trace in a file named as the argument of `--vcd`. The user still needs to give an argument to the `--vcd` switch, which will be irgnored.

kroening:
> Thanks for your feedback. @kroening
> 
> Do we want to keep the argument for VCD filename?
> 
> If yes, I can include the filename argument as a suffix in the final filename.

This is what we are doing for `--random-traces` (look at random_traces.cpp, line 169).

However, making that change would still regress the current behaviour. I am not opposed to this in principle, but we'd need to do this as part of a major version release. Let's do it.

src/ebmc/report_results.cpp
```diff
@@ -109,7 +109,7 @@ void report_results(
 
       switch(property.status)
       {
-      // clang-format off
+        // clang-format off
```
tautschnig: Please avoid whitespace changes.

src/ebmc/report_results.cpp
```diff
@@ -139,9 +139,8 @@ void report_results(
 
       if(property.has_witness_trace())
       {
-        auto term = [&property]() {
-          return property.is_exists_path() ? "Trace" : "Counterexample";
-        };
+        auto term = [&property]()
+        { return property.is_exists_path() ? "Trace" : "Counterexample"; };
```
tautschnig: As above: no whitespace changes, please.

src/ebmc/report_results.cpp
ShashankVM: Thanks for your review. These got added when I ran formatting in clang. Looks like I messed up, reverted the changes now.

src/ebmc/report_results.cpp
ShashankVM: Sure, will avoid whitespace changes going forward.

src/ebmc/report_results.cpp
```diff
@@ -176,18 +176,32 @@ void report_results(
 
   if(cmdline.isset("vcd"))
   {
+    const auto outfile_prefix = [&cmdline]() -> std::optional<std::string> {
+    if(cmdline.isset("vcd"))
```
kroening: This will always be true in this branch -- look at the if in line 177.

kroening:
Please apply `git-clang-format`.

src/ebmc/report_results.cpp
ShashankVM: Thanks for spotting this, fixed it and verified locally.

src/ebmc/report_results.cpp
```diff
@@ -176,18 +176,29 @@ void report_results(
 
   if(cmdline.isset("vcd"))
   {
+    const auto outfile_prefix = [&cmdline]() -> std::optional<std::string> {
```
kroening: No need to return an std::optional -- at this point, you always have a value, since --vcd is given. Simply return std::string.

kroening: Or even simpler:
```const auto outfile_prefix = cmdline.get_value("vcd") + '.';```

File: src/ebmc/report_results.cpp [hunk 5]
```diff
@@ -176,18 +176,29 @@ void report_results(
...
+
     for(const auto &property : result.properties)
     {
       if(property.has_witness_trace())
       {
-        std::string vcdfile = cmdline.get_value("vcd");
-        auto outfile = output_filet{vcdfile};
-
+        std::stringstream vcdfile; 
+        vcdfile << property.name << "_witness.vcd";
```
kroening: Simply:
```const auto filename = outfile_prefix + id2string(property.name) + "_witness.vcd";```

File: src/ebmc/report_results.cpp [hunk 6]
```diff
@@ -176,18 +176,29 @@ void report_results(
...
+        std::string filename;
+        if (outfile_prefix.has_value()) {
+          filename = outfile_prefix.value() + vcdfile.str();
+        } else {
+          filename = vcdfile.str();
+        }
+        
+        auto outfile = output_filet{filename};
+        std::cout << "Writing witness trace VCD file to " << filename
```
kroening: ```message.status() << "Writing ... " ... << messaget::eom;```

src/ebmc/report_results.cpp
```diff
@@ -176,18 +176,19 @@ void report_results(
...
       {
-        std::string vcdfile = cmdline.get_value("vcd");
-        auto outfile = output_filet{vcdfile};
-
+        const auto filename =
+          outfile_prefix + id2string(property.name) + "_witness.vcd";
+        auto outfile = output_filet{filename};
         messaget message(message_handler);
+        message.status() << "Writing witness trace VCD file to " << filename
+                         << "\n";
```
kroening: Make that `messaget::eom` instead of `'\n'` and we are there!

PR #1302, opened 2025-10-08
Verilog: use `zero_extend_exprt`

kroening:
This replaces a custom encoding via `typecast_exprt` used in the Verilog front-end by `zero_extend_exprt`.

### DIFF
--- a/regression/ebmc/smv-word-level/verilog6.desc
+++ b/regression/ebmc/smv-word-level/verilog6.desc
@@ -1,0 +1,7 @@
+CORE
+verilog6.sv
+--smv-word-level
+^CTLSPEC 0ud8_1 \+ 0ub6_000000 :: -0sd2_1 = 0ud8_4$
+^EXIT=0$
+^SIGNAL=0$
+--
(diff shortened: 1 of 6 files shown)


src/verilog/verilog_lowering.cpp
```diff
@@ -625,6 +625,36 @@ exprt verilog_lowering(exprt expr)
     else
       return expr;
   }
+  else if(expr.id() == ID_zero_extend)
```
tautschnig: So why is this lowering required? Shouldn't the back-end be capable of handling zero_extend?

kroening: The back-end is missing some cases; I'll separate.

kroening: The non-four-valued cases now work without lowering after updating to cbmc 5.7.1.

tautschnig: Will the four-valued cases be covered in a separate PR?

kroening: Yes, will do separate PR

PR #1322, opened 2025-10-15
Verilog: KNOWNBUG test for unconnected module input ports

kroening:


regression/verilog/modules/unconnected_ports1.desc
```diff
@@ -0,0 +1,6 @@
+KNOWNBUG
+unconnected_ports1.v
+--bound 0
+^EXIT=0$
+^SIGNAL=0$
+--
```
tautschnig, line 6: Would you mind adding a sentence stating what is going wrong here?

kroening: Done

tautschnig (approved):

PR #1331, opened 2025-10-17
SMV: add enums to the parse tree

kroening:
This adds enums as items to the parse tree data structure.


src/smvlang/parser.y
```diff
@@ -676,6 +676,11 @@ enum_element: IDENTIFIER_Token
            {
              $$=$1;
              PARSER.module->enum_set.insert(stack_expr($1).id_string());
+
+             exprt expr(ID_symbol);
```
tautschnig, line 680: Why is this not using symbol_exprt?

PR #1360, opened 2025-10-23
Verilog: clean up expression construction

kroening:
This replaces instances of manually constructed expressions by their proper constructors in the Verilog synthesis code.

### DIFF
--- a/src/verilog/verilog_synthesis.cpp
+++ b/src/verilog/verilog_synthesis.cpp
@@ -555,9 +555,7 @@ void verilog_synthesist::assignment_rec(
 
       if(it->type().id()==ID_bool)
       {
-        exprt bit_extract(ID_extractbit, it->type());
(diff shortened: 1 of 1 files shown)


src/verilog/verilog_synthesis.cpp
```diff
@@ -555,9 +555,7 @@ void verilog_synthesist::assignment_rec(
 
       if(it->type().id()==ID_bool)
       {
-        exprt bit_extract(ID_extractbit, it->type());
-        bit_extract.add_to_operands(rhs);
-        bit_extract.add_to_operands(offset_constant);
+        exprt bit_extract = extractbit_exprt{rhs, offset_constant};
```
tautschnig, line 558: Any reason not to use
```suggestion
        extractbit_exprt bit_extract{rhs, offset_constant};
```
instead? Equally applies to all changes below.

PR #1362, opened 2025-10-26
SMV: introduce `smv_identifier_exprt`

kroening:
This introduces a class for the identifiers used in the SMV parse tree.

The type checker subsequently turns them into symbols, when this is appropriate.


src/smvlang/parser.y
```diff
@@ -142,8 +142,8 @@ Function: merge_complex_identifier
 
 irep_idt merge_complex_identifier(const exprt &expr)
 {
-  if(expr.id() == ID_symbol)
-    return to_symbol_expr(expr).get_identifier();
+  if(expr.id() == ID_smv_identifier)
+    return expr.get(ID_identifier);
```
tautschnig: Should probably be `to_smv_identifier_expr(expr).identifier()`

PR #1368, opened 2025-10-28
better error reporting for Buechi conversion

kroening:
This improves the error reporting when the Buechi conversion is given an unsupported property.

### DIFF
--- a/src/ebmc/instrument_buechi.cpp
+++ b/src/ebmc/instrument_buechi.cpp
@@ -25,9 +25,7 @@ void instrument_buechi(
       continue;
 
     // This is for LTL and some fragment of SVA only.
-    if(
-      !is_LTL(property.normalized_expr) &&
-      !is_Buechi_SVA(property.normalized_expr))
+    if(!is_LTL(property.normalized_expr) && !is_SVA(property.normalized_expr))
     {
       property.unsupported("not convertible to Buechi");
       continue;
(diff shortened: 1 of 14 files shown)


src/ebmc/instrument_buechi.cpp
```diff
@@ -37,65 +35,74 @@ void instrument_buechi(
...
     {
-      // Note that we have negated the property,
-      // so this is the negation of the Buechi acceptance condition.
-      property_conjuncts.push_back(
-        F_exprt{G_exprt{not_exprt{buechi.liveness_signal}}});
-      have_liveness = true;
+      // make the automaton for the negation of the property
+      auto buechi =
+        ltl_to_buechi(not_exprt{property.normalized_expr}, message_handler);
+
```
tautschnig, line 43: I'm not entirely comfortable putting a large block of code into a `try` clause as it makes it far from clear which statements can raise an exception. Am I wrong in that `ltl_to_buechi` is actually the only call in here that can result in the specific exception that's later caught? I acknowledge that this is just me being uncomfortable, it's not a hard blocker.

tautschnig (approved):

src/ebmc/instrument_buechi.cpp
kroening: Yes, it is; this bugged me as well. To fix this, you'd need to extract that block into a function.

PR #1372, opened 2025-10-29
CI: speed up Ubuntu runs

kroening:
This is an attempt to speed up CI jobs on Ubuntu runners, which spend between 2 and 3 minutes on `apt-get update` and `apt-get install`.


kroening:
The question is whether Github would update the image often enough?

tautschnig (approved):

PR #1387, opened 2025-11-04
Homebrew formulas for ebmc 5.7 and 5.8

kroening:


tautschnig (approved):
Is it worth picking up some ideas from https://github.com/diffblue/homebrew-cbmc/blob/main/.github/workflows/add-formula.yml? Reviewing these formulas is actually not super fun (the only interesting piece is the precise commit id in each formula), so I'd rather have them created automatically.

PR #1389, opened 2025-11-05
SMV: word constants

kroening:
This adds NuSMV's word constants to scanner, parser, type checker.


src/smvlang/smv_typecheck.cpp
```diff
@@ -582,6 +583,126 @@ void smv_typecheckt::typecheck(exprt &expr, modet mode)
...
+  }
+
+  auto digits_int = string2integer(digits, base);
+
+  auto type =
+    bitvector_typet{is_signed ? ID_signedbv : ID_unsignedbv, bits.value()};
+
+  auto result = from_integer(digits_int, type);
+  result.add_source_location() = src_expr.source_location();
+  return result;
```
tautschnig:
```suggestion
  return from_integer(digits_int, type).with_source_location(src_expr.source_location());
```

kroening: Unfortunately won't work since we are returning a `constant_exprt`.

tautschnig: Then we should add `with_source_location` to `constant_exprt` (as we already have for some others)! (Obviously not a blocker for this PR here).

kroening: Changed return type to `exprt`

PR #1398, opened 2025-11-09
SMV: submodule type

kroening:
This introduces a class to document the existing SMV `submodule` type.


src/smvlang/smv_types.h
```diff
@@ -71,4 +71,35 @@ inline smv_enumeration_typet &to_smv_enumeration_type(typet &type)
   return static_cast<smv_enumeration_typet &>(type);
 }
 
+class smv_submodule_typet : public typet
```
tautschnig, line 75: Not that our documentation is particularly rich in general, but I'd appreciate some Doxygen here. It seems that coding agents benefit quite a lot from such documentation, so I think such (time) investment is worthwhile.

tautschnig (approved):

PR #1402, opened 2025-11-10
SMV: remove `vars` member in parse tree

kroening:
The variables are now stored as items, in the order in the source file, and this auxiliary member is no longer needed.


tautschnig (approved):
This seems to fix the test `var/already_declared4.desc`, so that test should move from `KNOWNBUG` to `CORE`.

PR #1404, opened 2025-11-10
SMV: catch variables that have the same identifier as an enum

kroening:
This errors variable declarations that use the same identifier as a previously defined enum.


tautschnig:
clang-format is asking for whitespace changes.

PR #1405, opened 2025-11-10
SMV: catch variables that have the same identifier as a parameter

kroening:
This errors variable and enum declarations that use the same identifier as a previously defined module parameter.

### DIFF
--- a/src/smvlang/smv_typecheck.cpp
+++ b/src/smvlang/smv_typecheck.cpp
@@ -60,7 +60,9 @@ public:
   } modet;
 
   void convert(smv_parse_treet::modulet &);
-  void create_var_symbols(const smv_parse_treet::modulet::item_listt &);
+  void create_var_symbols(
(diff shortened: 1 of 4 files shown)


src/smvlang/smv_typecheck.cpp
```diff
@@ -2133,13 +2135,20 @@ Function: smv_typecheckt::create_var_symbols
...
+  const smv_parse_treet::modulet::item_listt &items,
+  const std::list<irep_idt> &module_parameters)
 {
   const irep_idt mode = "SMV";
 
   // to catch variables that have the same name as enums
   std::unordered_set<irep_idt, irep_id_hash> enums;
 
+  // to catch re-use of parameter identifiers
+  std::unordered_set<irep_idt> parameters;
```
tautschnig: I'd say both or none should use `, irep_id_hash`.

tautschnig (approved):

PR #1411, opened 2025-11-13
Verilog: KNOWNBUG test for port connection that has typedef name

kroening:


regression/verilog/modules/port_connection1.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+port_connection1.sv
+
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
```
tautschnig, line 7: I'd appreciate a note saying what is going wrong here.

tautschnig (approved):

PR #1414, opened 2025-11-15
Make the counter synthesizable, update comment

ShashankVM:
- Use nonblocking assignment for counter. Sequential design use nonblocking assignments.
- Add reset and remove initial block as initial block is not synthesizable
- Add counter output port
-  Update comment to count up


ShashankVM:
Regression fails with error: unsupported property - only SVA s_eventually or AF implemented
Looks like the syntax is correct and not supported by the tool.

PR #1415, opened 2025-11-17
KNOWNBUG test for parameter without default value

kroening:
SystemVerilog 1800-2017 allows module parameter ports without default value.

### DIFF
--- a/regression/verilog/modules/parameter_without_default1.desc
+++ b/regression/verilog/modules/parameter_without_default1.desc
@@ -1,0 +1,9 @@
+KNOWNBUG
+parameter_without_default1.sv
+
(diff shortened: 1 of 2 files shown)


regression/verilog/modules/parameter_ports2.v
```diff
@@ -6,6 +6,6 @@ endmodule
 
 module main;
 
-  sub #(123) submodule();
+  sub #(8'd123) submodule();
```
tautschnig: How is this change related to the KNOWNBUG test?

kroening: Unrelated, dropping

PR #1417, opened 2025-11-17
copy language_file.cpp and language_file.h from CBMC

kroening:
The logic implemented in CBMC's `language_filest` class typechecks modules starting from the leaves of the dependency tree.  This is an ill-fit for both SMV and Verilog, which expect elaboration and type checking to start from the root (or "main") module.

This copies the file that contains the `langauge_filet` and `language_filest` classes in order to make this change locally.


kroening:
> Is a next step to clean up some module-related references in the CBMC code base?

Yes, will clean out.

PR #1436, opened 2025-11-24
SystemVerilog: sequence and property ports

kroening:
This adds SV sequence and property ports.


kroening:
Closing; done in https://github.com/diffblue/hw-cbmc/pull/1923

PR #1438, opened 2025-11-25
SVA: sequence and property instance expressions

kroening:
IEEE 1800-2017 refers to sequence and property instances, and rewriting algorithms for them (F.4.1).

This introduces an expression class each for sequence and property instance expressions.


src/verilog/expr2verilog.cpp
```diff
@@ -1493,6 +1493,45 @@ Function: expr2verilogt::convert_value_range
 
 \*******************************************************************/
 
+expr2verilogt::resultt expr2verilogt::convert_sequence_property_instance(
```
tautschnig, line 1496: The comment block above now no longer matches the function.

PR #1449, opened 2025-12-01
Verilog: fix integral vs. bit vector terminology

kroening:
This changes the naming of methods and error messages.  The term 'integral' includes non-integer types, e.g., time variables.  Many operators require operands that are integral but exclude time variables.  The standard calls these "bit vector types".

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -426,7 +426,7 @@ void verilog_typecheck_exprt::no_bool_ops(exprt &expr)
 
 /*******************************************************************\
 
-Function: verilog_typecheck_exprt::must_be_integral
+Function: verilog_typecheck_exprt::must_be_bit_vector
 
   Inputs:
 
@@ -436,9 +436,9 @@ Function: verilog_typecheck_exprt::must_be_integral
 
 \*******************************************************************/
 
-void verilog_typecheck_exprt::must_be_integral(const exprt &expr)
+void verilog_typecheck_exprt::must_be_bit_vector(exprt &expr)
 {
-  // Throw an error if the given expression doesn't have an integral type.
+  // Throw an error if the given expression doesn't have a bitvector type.
   const auto &type = expr.type();
   if(type.id() == ID_bool)
   {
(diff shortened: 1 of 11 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -3157,9 +3154,8 @@ exprt verilog_typecheck_exprt::convert_binary_expr(binary_exprt expr)
     // logical right shift >>
     convert_expr(expr.lhs());
     convert_expr(expr.rhs());
-    must_be_integral(expr.lhs());
-    must_be_integral(expr.rhs());
-    no_bool_ops(expr);
```
tautschnig, line 3162: Some change broke two of the regression test, and the omission of `no_bool_ops` here seems among the most substantial changes. Or is it the removal around line 3088?

kroening: Now fixed!

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -3069,25 +3070,34 @@ exprt verilog_typecheck_exprt::convert_binary_expr(binary_exprt expr)
...
   }
   else if(expr.id()==ID_verilog_case_equality ||
           expr.id()==ID_verilog_case_inequality)
   {
     // === and !==
+    // Take any operand types except real and shortreal (1800-2017 Table 11-1).
     // The result is always Boolean, and semantically
     // a proper equality is performed.
-    expr.type()=bool_typet();
+    expr.type()=bool_typet{};
```
tautschnig: Add whitespace or leave unchanged entirely to keep clang-format happy.

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -3069,25 +3070,34 @@ exprt verilog_typecheck_exprt::convert_binary_expr(binary_exprt expr)
...
 
     convert_relation(expr);
 
-    // integral operands only
-    must_be_integral(expr.lhs());
-    must_be_integral(expr.rhs());
+    if(expr.lhs().type().id() == ID_verilog_real || expr.lhs().type().id() == ID_verilog_shortreal)
```
tautschnig: Line too long

PR #1454, opened 2025-12-02
ebmc language interface implementation for Verilog

kroening:
This adds an implementation of the `ebmc_languaget` interface for Verilog.


src/verilog/verilog_ebmc_language.cpp
```diff
@@ -0,0 +1,288 @@
...
+static bool parse(
+  const cmdlinet &cmdline,
+  ebmc_language_filest &language_files,
+  message_handlert &message_handler)
+{
+  for(unsigned i = 0; i < cmdline.args.size(); i++)
+  {
+    if(parse(cmdline, cmdline.args[i], language_files, message_handler))
+      return true;
+  }
```
tautschnig: Code modernisation: use ranged for

src/verilog/verilog_ebmc_language.h
```diff
@@ -0,0 +1,31 @@
...
+
+#include <ebmc/ebmc_language.h>
+
+class verilog_parse_treet;
+
+class verilog_ebmc_languaget : public ebmc_languaget
+{
+public:
+  verilog_ebmc_languaget(cmdlinet &_cmdline, message_handlert &_message_handler)
+    : ebmc_languaget(_cmdline, _message_handler)
```
tautschnig: Shouldn't this be `const cmdlinet &`?

src/verilog/verilog_ebmc_language.h
```diff
@@ -0,0 +1,31 @@
...
+  {
+  }
+
+  // produce the transition system, and return it
+  std::optional<transition_systemt> transition_system() override;
```
tautschnig, line 30: Should likely be marked `nodiscard`

src/verilog/verilog_ebmc_language.cpp
```diff
@@ -0,0 +1,288 @@
...
+  //
+  // preprocessing
+  //
+  if(cmdline.isset("preprocess"))
+  {
+    preprocess(cmdline, message_handler);
+    return {};
+  }
+
+  transition_systemt transition_system;
```
tautschnig: I'd move that declaration closer to where it will actually be used.

src/verilog/verilog_ebmc_language.cpp
```diff
@@ -0,0 +1,288 @@
...
+  const cmdlinet &cmdline,
+  message_handlert &message_handler,
+  transition_systemt &transition_system)
+{
+  std::string top_module;
+
+  if(cmdline.isset("module"))
+    top_module = cmdline.get_value("module");
+  else if(cmdline.isset("top"))
+    top_module = cmdline.get_value("top");
```
tautschnig, line 152: Is it ok if `top_module` remains unset?

src/verilog/verilog_ebmc_language.cpp
```diff
@@ -0,0 +1,288 @@
...
+  if(cmdline.isset('I'))
+    options.set_option("I", cmdline.get_values('I'));
+
+  options.set_option("force-systemverilog", cmdline.isset("systemverilog"));
+
+  // do -D
+  if(cmdline.isset('D'))
+    options.set_option("defines", cmdline.get_values('D'));
+
+  language->set_language_options(options, message_handler);
```
tautschnig, line 64: We have rather similar code in `parse` below, perhaps this should be factored out.

tautschnig (approved):

src/verilog/verilog_ebmc_language.h
kroening: Yes, but will require change to base class; will do with separate PR.

src/verilog/verilog_ebmc_language.h
kroening: Likewise, will go into base class.

src/verilog/verilog_ebmc_language.cpp
kroening: Yes, `get_module` handles that.

src/verilog/verilog_ebmc_language.cpp
kroening: Yes, will do separately.

PR #1466, opened 2025-12-07
Verilog: use resolver for typedefs

kroening:
Verilog typedefs need to be resolved by the type checker, to account for parameter-dependent prefixes.


tautschnig:
Is the still-in-draft #1459 the second commit in this PR? Should #1459 be ready-for-review?

kroening:
> Is the still-in-draft #1459 the second commit in this PR? Should #1459 be ready-for-review?

Yes, done!

PR #1468, opened 2025-12-10
KNOWNBUG test for SMV's `self`

kroening:


regression/smv/modules/self1.desc
```diff
@@ -0,0 +1,8 @@
+KNOWNBUG
+self1.smv
+^EXIT=0$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
+This does not parse.
```
tautschnig, line 9: The KNOWNBUG CI job suggests that that's not the case - it seems the test passes just fine!?

kroening: I am getting
```
file self1.smv line 4: syntax error, unexpected self before 'self'
```

regression/smv/modules/self1.desc
```diff
@@ -1,5 +1,5 @@
 KNOWNBUG
-process1.smv
+self1.smv
 
 ^EXIT=0$
```
tautschnig, line 4: The missing blank line was the root cause of the test spuriously passing in CI. Fixed.

PR #1487, opened 2025-12-15
netlist: split up `convert_trans_to_netlistt::map_vars`

kroening:
This splits `convert_trans_to_netlistt::map_vars` into two pieces: one that identifies the variables to be put into the var_map, and one that allocates net-list nodes for those variables.

### DIFF
--- a/src/trans-netlist/trans_to_netlist.cpp
+++ b/src/trans-netlist/trans_to_netlist.cpp
@@ -149,9 +149,9 @@ protected:
 
   std::optional<exprt> convert_property(const exprt &);
 
-  void map_vars(
-    const irep_idt &module,
-    netlistt &dest);
+  var_mapt build_var_map(const irep_idt &module);
+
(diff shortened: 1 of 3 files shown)


src/trans-netlist/trans_to_netlist.cpp
```diff
@@ -248,39 +248,53 @@ void convert_trans_to_netlistt::map_vars(
...
-    [](const symbolt *a, const symbolt *b)
-    { return a->name.compare(b->name) < 0; });
+ Purpose:
+
+\*******************************************************************/
+
+void convert_trans_to_netlistt::allocate_state_variables(netlistt &dest)
+{
+  boolbv_widtht boolbv_width(ns);
+
```
tautschnig: This is never used?!

kroening: Removed

PR #1488, opened 2025-12-15
`aig_nodet` constructors

kroening:
This adds constructors for the two kinds of AIG nodes, preventing uninitialized nodes.

### DIFF
--- a/src/trans-netlist/aig.h
+++ b/src/trans-netlist/aig.h
@@ -22,18 +22,30 @@ class aig_nodet {
 public:
   literalt a, b;
 
-  aig_nodet() {}
-
-  bool is_and() const { return a.var_no() != literalt::unused_var_no(); }
-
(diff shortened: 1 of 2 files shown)


src/trans-netlist/aig.h
```diff
@@ -22,18 +22,31 @@ class aig_nodet {
 public:
   literalt a, b;
 
-  aig_nodet() {}
+  explicit aig_nodet(literalt::var_not var_no)
+  {
+    a.set(literalt::unused_var_no(), false);
+    b.set(var_no, false);
```
tautschnig:
```suggestion
  explicit aig_nodet(literalt::var_not var_no)
    : a(literalt::unused_var_no(), false), b(var_no, false)
  {
```
To avoid unnecessary default-construction of `a`, `b` (before `.set` is called)

kroening: Fixed as proposed

PR #1501, opened 2025-12-20
Verilog: use `instance_array` for width of primitive gates

kroening:
This switches the `range` annotation on primitive gates to the unpacked array as required by 1800-2017.

### DIFF
--- a/src/verilog/verilog_typecheck.cpp
+++ b/src/verilog/verilog_typecheck.cpp
@@ -223,28 +223,15 @@ Function: verilog_typecheckt::typecheck_builtin_port_connections
 void verilog_typecheckt::typecheck_builtin_port_connections(
   verilog_inst_baset::instancet &inst)
 {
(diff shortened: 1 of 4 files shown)


src/verilog/verilog_typecheck.cpp
```diff
@@ -206,28 +206,15 @@ Function: verilog_typecheckt::typecheck_builtin_port_connections
...
+    inst.type() = bool_typet{};
   else
   {
-    type.id(ID_unsignedbv);
-    type.set(ID_width, integer2string(width));
+    // We'll turn a one-dimensional array into a bit-vector
+    auto &array_type = to_array_type(inst.instance_array());
+    auto width =
+      numeric_cast_v<mp_integer>(to_constant_expr(array_type.size()));
+    inst.type() = unsignedbv_typet{width};
```
tautschnig, line 234: I understand that this isn't any different from what it was before, but still: why is `unsignedbv_typet` the right type here? I would have picked `bv_typet`.

PR #1533, opened 2025-12-28
SMV: further support for assignments of sets to scalars

kroening:
This extends the support for assignments of sets to scalars in the SMV type checker, via recursive conversion to a predicate.


tautschnig:
Looks like this makes `expressions/smv_set4.desc` pass - is this expected?

PR #1549, opened 2025-12-28
Verilog: add instance names to the current scope

kroening:
This changes the parser to add module instance names to the current scope, so they become usable in case there is a typedef of the same name in a higher scope.

### DIFF
--- a/src/verilog/parser.y
+++ b/src/verilog/parser.y
@@ -3134,7 +3134,9 @@ name_of_gate_instance_opt:
 name_of_gate_instance:
           any_identifier unpacked_dimension_brace
                 { init($$, ID_inst);
-                  stack_expr($$).set(ID_base_name, stack_expr($1).get(ID_base_name));
+                  auto base_name = stack_expr($1).get(ID_base_name);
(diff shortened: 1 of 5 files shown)


src/verilog/parser.y
```diff
@@ -3131,15 +3131,17 @@ name_of_gate_instance_opt:
...
+		  stack_expr($$).set(ID_base_name, base_name);
+		  PARSER.scopes.add_name(base_name, "", verilog_scopet::MODULE_INSTANCE);
 	          if(stack_expr($2).is_not_nil())
 	          {
 	            auto &range = stack_expr($$).add(ID_range);
 	            range = stack_expr($2).find(ID_range);
 	            range.id(ID_range);
 	          }
-	        }
+		}
```
tautschnig: May I suggest to consistently fix the indent? This indented `if` is confusing.

tautschnig (approved):

src/verilog/parser.y
kroening: #1554

PR #1561, opened 2026-01-03
Verilog: elaboration per compilation unit

kroening:
IEEE 1800 2017 3.12.1 requires compliant tools to compile "compilation units" separately, with a compilation unit scope.

This breaks out a function that elaborates a compilation unit by itself.

### DIFF
--- a/src/verilog/verilog_elaborate_compilation_unit.cpp
+++ b/src/verilog/verilog_elaborate_compilation_unit.cpp
@@ -1,0 +1,33 @@
+/*******************************************************************\
+
+Module: Elaboration of Verilog Compilation Units
+
+Author: Daniel Kroening, dkr@amazon.com
+
+\*******************************************************************/
+
+#include "verilog_elaborate_compilation_unit.h"
+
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_elaborate_compilation_unit.cpp
```diff
@@ -0,0 +1,33 @@
...
+\*******************************************************************/
+
+#include "verilog_elaborate_compilation_unit.h"
+
+#include "verilog_typecheck.h"
+
+void verilog_elaborate_compilation_unit(
+  const verilog_parse_treet &parse_tree,
+  symbol_table_baset &symbol_table,
+  message_handlert &message_handler)
```
tautschnig, line 16: There doesn't seem to be a need for a `message_handlert` here?!

kroening: It will come -- next, this will elaborate compilation-unit scoped typedefs, functions, etc.

PR #1575, opened 2026-01-07
KNOWNBUGs for loop index declaration and SMV conversion

Romy15200:
This PR adds three testcases covering:
1. Loop variable declaration in a for-loop header
2. Genvar declaration in a generate-for header
3. A case where using --smv-netlist together with --buechi produces an SMV that, when translated to AIG, yields a provable property

### DIFF
--- a/regression/ebmc/smv-netlist/array_assignment.desc
+++ b/regression/ebmc/smv-netlist/array_assignment.desc
@@ -1,0 +1,10 @@
+KNOWNBUG
+array_assignment.sv
+--reset rst --smv-netlist 
+
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+Outputs an SMV with an undeclared nondet. When declaring the nondet, nuXmv falsifies the property depsite it being correct.
+
--- a/regression/verilog/for/for_variable_declaration.desc
+++ b/regression/verilog/for/for_variable_declaration.desc
@@ -1,0 +1,10 @@
+KNOWNBUG
+for_variable_declaration.sv
+
+^EXIT=10$
+^SIGNAL=0$
+^no properties
+--
+--
+Declaration of loop index in for-loop header is not recognized
+
--- a/regression/verilog/for/for_variable_declaration.sv
+++ b/regression/verilog/for/for_variable_declaration.sv
@@ -1,0 +1,16 @@
+module main;
+
+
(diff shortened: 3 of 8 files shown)


regression/verilog/for/for_variable_declaration.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+for_variable_declaration.sv
+^EXIT=10$
```
tautschnig, line 4: The third line of a .desc file is for command-line options. Please leave it blank if no additional options are required. See `regression/test.pl --help` for an explanation of the file format.

regression/verilog/for/for_variable_declaration.desc
```diff
@@ -0,0 +1,7 @@
...
+^SIGNAL=0$
+^no properties
+--
+
```
tautschnig, line 10: After another `--` line please include a brief note stating what (undesired) behaviour/result you are seeing right now.

regression/verilog/for/for_variable_declaration.sv
```diff
@@ -0,0 +1,17 @@
+module main;
+
+
+logic [1:0] some_logic;
+
+
+always_comb begin
+int i;
+    for (i=0; i<2; i++) begin     // i cannot be defined in the loop header
```
kroening: Change the test so that `i` is in the loop initialisation!

Romy15200: Sorry, just did

regression/ebmc/smv-netlist/array_assignment.desc
```diff
@@ -0,0 +1,10 @@
+KNOWNBUG
+array_assignment.sv
+--reset rst --smv-netlist 
+
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+Outputs an SMV with an undeclared nondet. When declaring the nondet, nuXmv falsifies the property depsite it being correct.
```
tautschnig, line 9: It looks like this test actually passes, which likely implies that we need some extra patterns to check for expected/unexpected output.

kroening: To make progress, I'll merge, and will do a separate PR to fix this.

PR #1579, opened 2026-01-08
SystemVerilog: multiple assignments for `for` initialization

kroening:
1800-2017 allows multiple assignments in the initialization part of `for` loops.

### DIFF
--- a/src/verilog/parser.y
+++ b/src/verilog/parser.y
@@ -3565,6 +3565,13 @@ list_of_net_assignments:
                 { $$=$1;    mto($$, $3); }
         ;
 
+list_of_variable_assignments:
+          variable_assignment
+                { init($$); mto($$, $1); }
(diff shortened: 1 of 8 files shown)


src/verilog/parser.y
```diff
@@ -4019,7 +4026,13 @@ loop_statement:
                 { init($$, ID_for); mto($$, $3); mto($$, $5); mto($$, $7); mto($$, $9); }
         ;
 
-for_initialization: blocking_assignment
+for_initialization:
+          list_of_variable_assignments
+                { // make these ID_verilog_blocking_assign
```
tautschnig: Is there any reason not to include this work in this PR?

kroening: apologies, unclear comment -- this happens in the lines below. Will reword.

PR #1583, opened 2026-01-12
simpler netlist generation

kroening:
This adds a variant of the `convert_trans_to_netlistt` implementation that is much simpler, but may yield larger netlists.  It is hence added as a command-line option `--simple-netlist`.


regression/verilog/synthesis/always_comb2.aig.desc
```diff
@@ -9,4 +9,3 @@ always_comb2.sv
 --
 ^warning: ignoring
 --
-This segfaults.
```
tautschnig, line 12: I suppose it would still segfault when not using `--simple-netlist`? If so, shouldn't we still keep a `KNOWNBUG` test as well?

kroening: Will do

PR #1586, opened 2026-01-12
`bv_varidt` now has `next` member

kroening:
This adds a `next` member to `bv_varidt`.

### DIFF
--- a/src/trans-netlist/bv_varid.h
+++ b/src/trans-netlist/bv_varid.h
@@ -18,22 +18,28 @@ class bv_varidt
 public:
   irep_idt id;
   std::size_t bit_nr;
+  bool next;
 
   friend bool operator==(const bv_varidt &i1, const bv_varidt &i2)
   {
-    return i1.id==i2.id && i1.bit_nr==i2.bit_nr;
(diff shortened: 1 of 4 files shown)


src/trans-netlist/bv_varid.h
```diff
@@ -47,7 +53,9 @@ static inline std::ostream & operator<< (std::ostream &out, const bv_varidt &v)
 struct bv_varidt_hash
 {
   size_t operator()(const bv_varidt &bv_varid) const
-  { return hash_string(bv_varid.id)^bv_varid.bit_nr; }
+  {
+    return hash_string(bv_varid.id) ^ bv_varid.bit_nr ^ bv_varid.next;
```
tautschnig, line 57: Should this perhaps use the hash function helpers from `irep_hash.h` instead of naively XOR-ing values? `bit_nr` will only ever have some of the least-significant bits set, and `next` will only ever be `0` or `1`. Alas, only `id` will make use of all bits, while the other two will only ever affect a small subset of the bits. In contrast, `hash_combine` would take care of rotating and distributing bits.

PR #1589, opened 2026-01-17
extract `convert_trans_to_netlistt::build_var_map` into separate file

kroening:
This extracts the method `convert_trans_to_netlistt::build_var_map` into a separate function in a separate file.

### DIFF
--- a/src/trans-netlist/build_netlist_var_map.h
+++ b/src/trans-netlist/build_netlist_var_map.h
@@ -1,0 +1,18 @@
+/*******************************************************************\
+
+Module: var_map for Netlist
+
+Author: Daniel Kroening, dkr@amazon.com
+
+\*******************************************************************/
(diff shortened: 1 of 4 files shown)


src/trans-netlist/build_netlist_var_map.h
```diff
@@ -0,0 +1,20 @@
...
+
+\*******************************************************************/
+
+#ifndef CPROVER_TRANS_NETLIST_BUILD_NETLIST_VAR_MAP_H
+#define CPROVER_TRANS_NETLIST_BUILD_NETLIST_VAR_MAP_H
+
+#include "netlist.h"
+
+//#include <util/mathematical_expr.h>
+//#include <util/namespace.h>
```
tautschnig: I think those should be removed.

PR #1590, opened 2026-01-18
SMV netlists: declare variables for the nondeterministic nodes

kroening:
This adds the missing declarations for nondeterministic nodes in netlists to the SMV output.


tautschnig (approved):
Shouldn't this fix `regression/ebmc/smv-netlist/array_assignment.desc` (though seemingly it does not, else CI would have complained)?

PR #1593, opened 2026-01-20
move `cnf_gate_and`

kroening:
The `cnf_gate_and` function has one user; this moves the implementation from a global header file to that user.

### DIFF
--- a/src/trans-netlist/unwind_netlist.cpp
+++ b/src/trans-netlist/unwind_netlist.cpp
@@ -21,6 +21,40 @@ Author: Daniel Kroening, kroening@kroening.com
 
 /*******************************************************************\
 
+Function: cnf_gate_and
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+// This is a copy of cnft::gate_and, which is protected.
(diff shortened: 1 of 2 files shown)


src/trans-netlist/unwind_netlist.cpp
```diff
@@ -21,6 +21,39 @@ Author: Daniel Kroening, kroening@kroening.com
...
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+inline void cnf_gate_and(cnft &cnf, literalt a, literalt b, literalt o)
```
tautschnig, line 35: Is the the protected-ness of `cnft::gate_and` the sole reason this function exists in the first place?

kroening: Yes.

tautschnig: May I ask for a comment stating this?

kroening: BTW, the interface in `cnft` has been the exact same since before our commit history started. It's low-risk to make these methods public.

tautschnig: https://github.com/diffblue/cbmc/pull/8827 will make them public.

PR #1596, opened 2026-01-21
add flag to Buechi tests

kroening:
This adds a flag to the tests that use `--buechi` to run them selectively.


tautschnig (approved):
I'd suggest adding a `Makefile` rule that runs those, even if just for documentation purposes.

kroening:
Coming!

PR #1599, opened 2026-01-22
tautology checker now only considers properties with unknown state

kroening:
The tautology checker should only be applied to properties that are still unknown.

### DIFF
--- a/src/ebmc/tautology_check.cpp
+++ b/src/ebmc/tautology_check.cpp
@@ -74,9 +74,13 @@ property_checker_resultt tautology_check(
 
   for(auto &property : result.properties)
   {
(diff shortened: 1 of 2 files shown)


src/ebmc/tautology_check.cpp
```diff
@@ -74,9 +74,12 @@ property_checker_resultt tautology_check(
 
   for(auto &property : result.properties)
   {
-    if(is_tautology(property.normalized_expr, solver_factory, message_handler))
+    if(property.is_unknown())
     {
-      property.proved("tautology");
+      if(is_tautology(property.normalized_expr, solver_factory, message_handler))
```
tautschnig: Line too long, wants to be wrapped.

PR #1604, opened 2026-01-26
netlist conversion: track equivalent nodes

kroening:
This replaces the list of constraints on netlist nodes (netlist nodes that have to evaluate to true) by a list of node equivalences.  This is more compact when node equivalences have to be generated, as the AIG nodes for lhs<->rhs do not have to be generated.

### DIFF
--- a/src/trans-netlist/aig.h
+++ b/src/trans-netlist/aig.h
@@ -154,14 +154,27 @@ protected:
 
 std::ostream &operator<<(std::ostream &, const aigt &);
 
-class aig_plus_constraintst : public aigt {
+class aig_plus_constraintst : public aigt
+{
 public:
   typedef std::vector<literalt> constraintst;
   constraintst constraints;
 
-  void clear() {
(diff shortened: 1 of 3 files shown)


src/ic3/r0ead_input.cc
```diff
@@ -217,12 +217,13 @@ void CompInfo::form_var_nums()
   ================================================*/
 void ic3_enginet::add_verilog_conv_constrs()
 {
+  PRECONDITION(netlist.equivalences.empty());
```
tautschnig: Why is this an important invariant to maintain for this function?

kroening: #1605 will remove these

tautschnig: But then, at present, #1605 does not _change_ any existing code, it just adds new code (and, hence, isn't actually being used).

kroening: Now using a different approach -- the equivalences are now recorded in addition to the constraints

PR #1608, opened 2026-01-27
netlists now come with AIG labels

kroening:
The netlist generation code now attaches labels based on the variable identifiers and bit number to the AIG nodes that represent them.


tautschnig (approved):
Do we have any tests of this code?

PR #1610, opened 2026-01-31
pretty-printing SMV for binary CTL operators

kroening:
Instead of printing `a AU b`, output `A[a U b]`, as expected by NuSMV.

### DIFF
--- a/src/smvlang/expr2smv.cpp
+++ b/src/smvlang/expr2smv.cpp
@@ -11,6 +11,8 @@ Author: Daniel Kroening, kroening@kroening.com
 
 #include <util/mathematical_types.h>
 
+#include <temporal-logic/ltl.h>
+
 #include "smv_expr.h"
 #include "smv_types.h"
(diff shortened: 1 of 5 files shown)


src/smvlang/expr2smv.cpp
```diff
@@ -134,6 +136,31 @@ expr2smvt::resultt expr2smvt::convert_binary(
...
+
+ Purpose:
+
+\*******************************************************************/
+
+expr2smvt::resultt expr2smvt::convert_binary_ctl(
+  const binary_exprt &src,
+  const std::string &symbol)
+{
+  PRECONDITION(symbol.size() == 2);
```
tautschnig, line 153: Should this be extended to test for specific characters even? Like `PRECONDITION(symbol[0] == 'A' || symbol[0] == 'E');`

kroening: added

PR #1618, opened 2026-02-02
CI: automatic PR for Homebrew formula

kroening:
This extends the ebmc release action to create a PR that adds the Homebrew formula.

### DIFF
--- a/.github/workflows/ebmc-release.yaml
+++ b/.github/workflows/ebmc-release.yaml
@@ -321,3 +321,71 @@ jobs:
           \`\`\`
           EOM
           gh release edit ebmc-${{ env.EBMC_VERSION }} --repo ${{ github.repository }} --draft=false --notes-file body
+
+  homebrew-formula-PR:
+    name: create PR for Homebrew formula
+    runs-on: ubuntu-24.04
+    needs: [get-version-information]
+    permissions:
+      contents: write
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: get on the main branch
+        run: |
+          git fetch origin main
+          git checkout main
+
+      - name: create Homebrew formula
+        env:
+          EBMC_VERSION: ${{ needs.get-version-information.outputs.version }}
+        run: |
+          cat << EOM > Formula/ebmc@${{ env.EBMC_VERSION }}.rb
+          class Ebmc < Formula
+            desc "Model Checker for SystemVerilog"
+            homepage "https://www.cprover.org/ebmc/"
+            url "https://github.com/diffblue/hw-cbmc.git",
+              tag: "ebmc-${{ env.EBMC_VERSION }}",
+              revision: "${{ github.sha }}"
+            version "ebmc-${{ env.EBMC_VERSION }}"
+            license "BSD-3-Clause"
+
+            uses_from_macos "flex" => :build
+            uses_from_macos "curl" => :build
+            depends_on "bison" => :build
+
+            def install
+              system "make", "-C", "lib/cbmc/src", "minisat2-download"
+              system "make", "-C", "src"
+              system "mkdir", "-p", "#{prefix}/usr/bin"
+              system "cp", "src/ebmc/ebmc", "#{prefix}/usr/bin/"
+            end
+
+            test do
+              system "make", "-C", "regression/ebmc", "test"
+            end
+          end
+          EOM
+
+      - name: create the commit
+        env:
+          EBMC_VERSION: ${{ needs.get-version-information.outputs.version }}
+        run: |
+          FILE=Formula/ebmc@${{ env.EBMC_VERSION }}.rb
+          git add $FILE
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git commit -m "Homebrew formula for ebmc ${{ env.EBMC_VERSION }}" $FILE
+
+      - name: create PR
+        env:
+          EBMC_VERSION: ${{ needs.get-version-information.outputs.version }}
+        uses: peter-evans/create-pull-request@v8
+        with:
+          branch: ebmc-formula-${{ env.EBMC_VERSION }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ebmc-release.yaml
```diff
@@ -321,3 +321,75 @@ jobs:
...
+      pull-requests: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: get on the main branch
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          git fetch origin main
+          git checkout main
```
tautschnig, line 338: Why is this necessary? Should this perhaps be replaced by code that checks that we are on `main`?

.github/workflows/ebmc-release.yaml
```diff
@@ -321,3 +321,75 @@ jobs:
...
+
+            uses_from_macos "flex" => :build
+            uses_from_macos "curl" => :build
+            depends_on "bison" => :build
+
+            def install
+              system "make", "-C", "lib/cbmc/src", "minisat2-download"
+              system "make", "-C", "src"
+              system "mkdir", "-p", "#{prefix}/usr/bin"
+              system "cp", "src/ebmc/ebmc", "#{prefix}/usr/bin/"
```
tautschnig, line 362: Not for this PR, but since it came up while reading this: should we also put other binaries in the package like `hw-cbmc` and `vlindex`? (This would apply to not just homebrew, the same question holds for all other packages.)

.github/workflows/ebmc-release.yaml
```diff
@@ -321,3 +321,75 @@ jobs:
...
+            end
+          end
+          EOM
+
+      - name: create and push branch
+        env:
+          EBMC_VERSION: ${{ needs.get-version-information.outputs.version }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          FILE=Formula/ebmc@${{ env.EBMC_VERSION }}.rb
```
tautschnig, line 375: I'd suggestion to put this into `$GITHUB_ENV` once rather than computing this file name multiple times.

.github/workflows/ebmc-release.yaml
```diff
@@ -321,3 +321,75 @@ jobs:
...
+          git add $FILE
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git commit -m "Homebrew formula for ebmc ${{ env.EBMC_VERSION }}" $FILE
+
```
tautschnig: Actually this entire piece seems surprising: `peter-evans/create-pull-request` happily takes care of this.

.github/workflows/ebmc-release.yaml
```diff
@@ -321,3 +321,75 @@ jobs:
...
+          EBMC_VERSION: ${{ needs.get-version-information.outputs.version }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        uses: peter-evans/create-pull-request@v8
+        with:
+          branch: ebmc-formula-${{ env.EBMC_VERSION }}
+          base: main
+          title: "Homebrew formula for ebmc ${{ env.EBMC_VERSION }}"
+          body: |
+            Automated update for ebmc Homebrew formula.
+            Created by .github/workflows/ebmc-release.yaml
```
tautschnig, line 391: You can get rid of the `create and push branch` step by adding `add-paths: Formula/ebmc@${{ env.EBMC_VERSION }}.rb` and `commit-message: Homebrew formula for ebmc ${{ env.EBMC_VERSION }}`. (And I'd add `delete-branch: true`)

.github/workflows/ebmc-release.yaml
```diff
@@ -321,3 +321,75 @@ jobs:
...
+          FILE=Formula/ebmc@${{ env.EBMC_VERSION }}.rb
+          git add $FILE
+          git config user.name "github-actions[bot]"
+          git config user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git commit -m "Homebrew formula for ebmc ${{ env.EBMC_VERSION }}" $FILE
+
+      - name: create PR
+        env:
+          EBMC_VERSION: ${{ needs.get-version-information.outputs.version }}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
tautschnig: Is that actually necessary here?

.github/workflows/ebmc-release.yaml
kroening: That is overly restrictive -- suppose you wanted to push a hot-fix for a particular release. Or someone goes ahead and merges something against the next release while the release CI is still running.

.github/workflows/ebmc-release.yaml
kroening: `hw-cbmc` isn't useful right now. Yes, we should start thinking about shipping `vlindex`!

.github/workflows/ebmc-release.yaml
kroening: Yes, will replace for next run.

.github/workflows/ebmc-release.yaml
kroening: I'll simply try.

.github/workflows/ebmc-release.yaml
tautschnig: But then this bit of code should really be removed - the `checkout` action will have taken care of checking out exactly the code that had been tagged (and caused this workflow to run).

.github/workflows/ebmc-release.yaml
kroening: They do not seem to be needed; removed.

.github/workflows/ebmc-release.yaml
kroening: Yes, and that will often be the same as `main`, but it doesn't need to be.

PR #1623, opened 2026-02-04
SMV: split type checking out of `convert_defines`

kroening:
This splits up the `convert_defines` method, to separate conversion from type checking.

### DIFF
--- a/src/smvlang/smv_typecheck.cpp
+++ b/src/smvlang/smv_typecheck.cpp
@@ -62,10 +62,6 @@ public:
 
   void create_var_symbols(const smv_parse_treet::modulet::element_listt &);
 
-  void collect_define(const exprt &lhs, const exprt &rhs);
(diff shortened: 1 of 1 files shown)


src/smvlang/smv_typecheck.cpp
```diff
@@ -151,6 +147,11 @@ class smv_typecheckt:public typecheckt
   typedef std::unordered_map<irep_idt, definet, irep_id_hash> define_mapt;
   define_mapt define_map;
 
+  void collect_define(const exprt &lhs, const exprt &rhs);
+  void typecheck_defines();
+  void typecheck_define(define_mapt::iterator);
```
tautschnig, line 152: For future consideration: I find passing around an iterator pretty opaque and would have preferred passing around two parameters: a (const) key `const irep_idt& identifier` and a (non-const) `definet &d`.

PR #1626, opened 2026-02-05
bump EBMC version to 5.10

kroening:
This bumps up the version number of ebmc post release of ebmc 5.9.


tautschnig (approved):
Perhaps it's also worth creating an initial changelog entry in this PR?

PR #1642, opened 2026-02-16
Verilog: non-wildcard import

kroening:
This adds 1800 2017 26.3 non-wildcard `import` declarations.

### DIFF
--- a/src/verilog/verilog_elaborate_type.cpp
+++ b/src/verilog/verilog_elaborate_type.cpp
@@ -343,7 +343,21 @@ typet verilog_typecheck_exprt::elaborate_type(const typet &src)
   {
     // Look it up!
     auto base_name = to_verilog_typedef_type(src).base_name();
-    const auto *symbol_ptr = resolve(base_name);
+    const symbolt *symbol_ptr;
+    auto import = src.get("import");
+    if(import != irep_idt{})
+    {
+      auto full_identifier = "Verilog::package::" + id2string(import);
+
+      if(ns.lookup(full_identifier, symbol_ptr))
+      {
+        DATA_INVARIANT(false, "failed to find imported typedef identifier");
+      }
+    }
+    else
+    {
+      symbol_ptr = resolve(base_name);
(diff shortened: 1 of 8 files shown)


src/verilog/verilog_elaborate_type.cpp
```diff
@@ -343,7 +343,22 @@ typet verilog_typecheck_exprt::elaborate_type(const typet &src)
   {
     // Look it up!
     auto base_name = to_verilog_typedef_type(src).base_name();
-    const auto *symbol_ptr = resolve(base_name);
+    const symbolt *symbol_ptr;
+
+    if(src.get("import") != irep_idt{})
+    {
+      auto full_identifier =
+        "Verilog::package::" + id2string(src.get("import"));
```
tautschnig: I'd suggest we do the `src.get("import")` only once: for both performance and maintenance reasons (there's a literal string in there).

src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1383,8 +1383,22 @@ exprt verilog_typecheck_exprt::convert_verilog_identifier(
   verilog_identifier_exprt expr,
   const std::optional<typet> &implicit_net_type)
 {
+  const symbolt *symbol;
   auto base_name = expr.base_name();
-  auto symbol = resolve(base_name);
+
+  if(expr.get("import") != irep_idt{})
+  {
+    auto full_identifier = "Verilog::package::" + id2string(expr.get("import"));
```
tautschnig: Just as above: do `expr.get("import")` only once.

tautschnig (approved):

src/verilog/verilog_elaborate_type.cpp
kroening: done

src/verilog/verilog_typecheck_expr.cpp
kroening: done

PR #1648, opened 2026-02-19
KNOWNBUG test for compilation-unit scoped parameter

kroening:


regression/verilog/parameters/compilation_unit_parameter1.desc
```diff
@@ -0,0 +1,7 @@
+KNOWNBUG
+compilation_unit_parameter1.sv
+
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
```
tautschnig, line 7: I'd appreciate a comment explaining what is going wrong here.

PR #1672, opened 2026-02-25
Verilog: comments and redundant `synth_expr` call

kroening:
This adds comments to the arguments of `assignment_rec`, which indicate that a redundant call to `synth_expr` can be removed.


tautschnig (approved):
Approving so as not the block all the work building on top of this, but I'd still prefer a solution not based on comments.

PR #1674, opened 2026-02-25
SystemVerilog: assignments to type casts

kroening:
SystemVerilog allows assignments to implicit conversions, e.g., from vectors to structs.


src/verilog/verilog_synthesis.cpp
```diff
@@ -31,6 +31,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <map>
 #include <set>
 
+exprt verilog_lowering_cast(typecast_exprt);
```
tautschnig: Can this please instead be made available via `verilog_lowering.h`?

tautschnig (approved):

PR #1683, opened 2026-02-28
SMV: use hex when the source constant has been given in hex

kroening:
The SMV pretty printer now uses base 16 when the original constant was given using Verilog's base 16 syntax.

### DIFF
--- a/src/smvlang/expr2smv.cpp
+++ b/src/smvlang/expr2smv.cpp
@@ -721,6 +721,19 @@ expr2smvt::resultt expr2smvt::convert_constant(const constant_exprt &src)
 {
   const typet &type = src.type();
 
+  auto use_hex = [](const exprt &src)
+  {
+    auto verilog_value = id2string(src.get("#verilog_value"));
+    auto tick_pos = verilog_value.find('\'');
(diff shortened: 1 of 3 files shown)


src/smvlang/expr2smv.cpp
```diff
@@ -721,6 +721,19 @@ expr2smvt::resultt expr2smvt::convert_constant(const constant_exprt &src)
 {
   const typet &type = src.type();
 
+  auto use_hex = [](const exprt &src)
+  {
+    auto verilog_value = id2string(src.get("#verilog_value"));
+    auto tick_pos = verilog_value.find('\'');
+    if(tick_pos != std::string::npos)
+    {
+      auto base_char = std::string(verilog_value, tick_pos + 1, 1);
```
tautschnig, line 730: Isn't that a very cumbersome way of doing `verilog_value[tick_pos + 1]`? (Though this yields a `char` instead of `std::string`, but that would be yet another improvement.)

kroening: I considered it, but then there'd need to be another guard for the length of the string.

PR #1693, opened 2026-03-03
SystemVerilog: detect import conflicts

kroening:
Per IEEE 1800-2017 section 26.3, this adds error detection for:
- Direct imports that conflict with local declarations
- Wildcard imports that are used in expressions and then redeclared

Wildcard imports that are not used remain 'potentially locally visible' and can be shadowed by local declarations.

### DIFF
--- a/regression/verilog/packages/import6.desc
+++ b/regression/verilog/packages/import6.desc
@@ -1,7 +1,7 @@
-KNOWNBUG
+CORE
 import6.sv
 --bound 0
-^EXIT=2$
+^EXIT=6$
 ^SIGNAL=0$
 --
 --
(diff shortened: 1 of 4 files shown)


regression/verilog/packages/import6.desc
```diff
@@ -1,7 +1,7 @@
-KNOWNBUG
+CORE
 import6.sv
 --bound 0
-^EXIT=2$
+^EXIT=6$
```
tautschnig, line 4: Here and below: can we also include a check for the presence of an error message?

regression/verilog/packages/import8.desc
```diff
@@ -1,7 +1,7 @@
-KNOWNBUG
+CORE
 import8.sv
 --bound 0
-^EXIT=2$
+^EXIT=6$
```
tautschnig, line 4: Same here: test for the error message.

PR #1696, opened 2026-03-05
SMV word-level output: map identifiers to SMV

kroening:
This adds a map from symbol table identifiers to syntactically valid SMV identifiers to the SMV word-level output.

### DIFF
--- a/src/ebmc/output_smv_word_level.cpp
+++ b/src/ebmc/output_smv_word_level.cpp
@@ -9,15 +9,44 @@ Author: Daniel Kroening, dkr@amazon.com
 #include "output_smv_word_level.h"
 
 #include <util/bitvector_types.h>
+#include <util/replace_symbol.h>
 
 #include <smvlang/expr2smv.h>
+#include <smvlang/smv_expr.h>
 #include <temporal-logic/sva_to_ltl.h>
+#include <trans-word-level/next_symbol.h>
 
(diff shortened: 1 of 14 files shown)


src/ebmc/output_smv_word_level.cpp
```diff
@@ -43,21 +78,73 @@ operator<<(std::ostream &out, const smv_type_printert &type_printer)
...
+
+    const symbolt &symbol = s_it->second;
+
+    if(is_supported_type(symbol.type))
+    {
+      auto smv_identifier = create_smv_identifier(
+        !symbol.base_name.empty() ? symbol.base_name : symbol.name);
+
+      smv_identifier_exprt replacement{smv_identifier, symbol.type};
+      result.insert(symbol.symbol_expr(), replacement);
```
tautschnig: Given that `create_smv_identifier` can produce the same SMV-compatible identifier for different input strings (owing to replacing various characters by `_`), shouldn't there be safeguards that `result` doesn't already replace something else by the same `smv_identifier`? `replace_symbolt` doesn't have this reverse map, so it would have to be maintained in here.

kroening: Yes, will add collision detection.

kroening: Now with collision detection.

PR #1703, opened 2026-03-08
EBMC: structural completeness threshold

kroening:
This adds a simple approximation of the recurrence diameter as completeness threshold for Gp and Fp.

### DIFF
--- a/regression/ebmc/engine-heuristic/counter4.desc
+++ b/regression/ebmc/engine-heuristic/counter4.desc
@@ -1,0 +1,10 @@
+CORE
+counter4.sv
+
+^\[main\.p0\] always main\.cnt != 4'b1111: REFUTED$
+^\[main\.p1\] always main\.cnt == main\.shadow: PROVED \(structural CT=255\)$
+^EXIT=10$
+^SIGNAL=0$
+--
+^warning: ignoring
+--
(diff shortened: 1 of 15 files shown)


src/ebmc/completeness_threshold.cpp
```diff
@@ -9,12 +9,68 @@ Author: Daniel Kroening, dkr@amazon.com
...
+    auto element_bits = state_bits(array_type.element_type());
+    auto size = numeric_cast<mp_integer>(array_type.size());
+    if(element_bits.has_value() && size.has_value())
+      return *element_bits * *size;
+    else
+      return {};
+  }
+  else
+    return {};
+}
```
tautschnig: This could just use `pointer_offset_bits`, which would yield exactly the same results.

regression/ebmc/engine-heuristic/counter4.desc
```diff
@@ -0,0 +1,10 @@
+CORE
+counter4.sv
+
+^\[main\.p0\] always main\.cnt != 4'b1111: REFUTED$
+^\[main\.p1\] always main\.cnt == main\.shadow: PROVED \(structural CT=255\)$
```
tautschnig, line 5: How is `structural CT=255` possible with the `if(total_bits > 8) return {};` code in `recurrence_diameter`?

kroening: CT 255 is exactly what you get for 8 bits!

src/ebmc/completeness_threshold.cpp
kroening: That doesn't know the verilog_*bv types.

PR #1732, opened 2026-03-14
Move lowering of bit select expressions to `verilog_lowering`

kroening:
This moves the lowering of Verilog's bit select expressions `a[b]` from the expression type checker to the lowering function.

### DIFF
--- a/src/verilog/verilog_lowering.cpp
+++ b/src/verilog/verilog_lowering.cpp
@@ -794,6 +794,68 @@ exprt verilog_lowering(exprt expr)
     else
       return expr;
   }
+  else if(expr.id() == ID_verilog_bit_select)
+  {
+    // This may be an array index or a bit extraction, depending
+    // on the type of the first operand.
+    auto &bit_select = to_verilog_bit_select_expr(expr);
+    auto &src = bit_select.src();
+
+    if(src.type().id() == ID_array)
+    {
+      // Lower to array index expression
+      auto &array_type = to_verilog_array_type(src.type());
+      auto index_type = array_type.index_type();
+      exprt index = typecast_exprt{bit_select.index(), index_type};
+
+      if(array_type.is_unpacked())
+      {
+        // For unpacked arrays, the internal representation stores
+        // elements starting from the left index of the range.
+        // We need to adjust the Verilog index to the internal index.
+        auto array_size = array_type.size_int();
+        auto offset = array_type.offset();
+
+        if(array_type.increasing())
+        {
+          // ascending range [l:r] with l<r, e.g., [0:4]
+          // internal index = verilog_index - offset
+          if(offset != 0)
+            index = minus_exprt{index, from_integer(offset, index_type)};
+        }
+        else
+        {
+          // descending range [l:r] with l>=r, e.g., [4:0]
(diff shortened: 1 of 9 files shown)


src/verilog/verilog_lowering.cpp
```diff
@@ -794,6 +794,72 @@ exprt verilog_lowering(exprt expr)
...
+      // Lower to array index expression
+      auto &array_type = to_array_type(src.type());
+      auto array_size_opt = numeric_cast<mp_integer>(array_type.size());
+      PRECONDITION(array_size_opt.has_value());
+      auto offset_expr = static_cast<const exprt &>(array_type.find(ID_offset));
+      auto offset_opt = numeric_cast<mp_integer>(offset_expr);
+      PRECONDITION(offset_opt.has_value());
+      auto _index_type = unsignedbv_typet(
+        address_bits((*array_size_opt + *offset_opt).to_ulong()));
+      auto index = typecast_exprt{bit_select.index(), _index_type};
```
tautschnig: Can we find a way to use `index_type()` from here, perhaps by moving `index_type` to some utility header? The above code is rather hard to read.

src/verilog/verilog_lowering.cpp
```diff
@@ -794,6 +794,72 @@ exprt verilog_lowering(exprt expr)
...
+        {
+          index = typecast_exprt{
+            minus_exprt{
+              minus_exprt{
+                plus_exprt{
+                  typecast_exprt{offset_expr, _index_type},
+                  typecast_exprt{array_type.size(), _index_type}},
+                from_integer(1, _index_type)},
+              index},
+            _index_type};
```
tautschnig: The outermost `typecast_exprt` seems unnecessary.

src/verilog/verilog_lowering.cpp
```diff
@@ -794,6 +794,72 @@ exprt verilog_lowering(exprt expr)
...
+        }
+      }
+
+      return index_exprt{src, index, array_type.element_type()};
+    }
+    else
+    {
+      // Lower to extractbit
+      auto width = verilog_bits(src.type());
+      auto offset = src.type().get_int(ID_C_offset);
```
tautschnig, line 846: Hmm, do we have `ID_offset` when an array and `ID_C_offset` otherwise? Seems a bit strange.

kroening: Yes, should be one or the other.

src/verilog/verilog_lowering.cpp
kroening: #1734

src/verilog/verilog_lowering.cpp
tautschnig: So do we actually have both of these in use?

kroening: Yes, it's `#offset` for bit vectors, and `offset` for the arrays.

tautschnig: What is the rationale for this?

kroening: This how these features were added incrementally -- I'll fix, but this should be done in a separate PR, not here.

PR #1741, opened 2026-03-19
Verilog: test for signed division overflow

kroening:
Test that `8'sh80 / 8'shff` yields `8'sh80`, exercising the case where the minimum representable 8-bit signed value (-128) is divided by -1. Per IEEE 1800-2017 section 6.9.1, arithmetic is modulo 2^n, so the result wraps back to -128.

### DIFF
--- a/regression/verilog/expressions/signed-division-overflow1.sv
+++ b/regression/verilog/expressions/signed-division-overflow1.sv
@@ -1,0 +1,9 @@
+module main(input signed [7:0] a, b);
+
+  // -128 / -1 overflows: +128 is not representable in 8 signed bits.
+  // Per 1800-2017 6.9.1, arithmetic is modulo 2^n, yielding -128.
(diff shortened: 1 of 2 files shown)


regression/verilog/expressions/signed-division-overflow1.sv
```diff
@@ -0,0 +1,7 @@
+module main;
+
+  // -128 / -1 overflows: +128 is not representable in 8 signed bits.
+  // Per 1800-2017 6.9.1, arithmetic is modulo 2^n, yielding -128.
+  initial assert (8'sh80 / 8'shff == 8'sh80);
```
tautschnig, line 5: Does this also work correctly when not using constants, i.e., do whether "works correctly" extends beyond the simplifier?

kroening: Added!

PR #1743, opened 2026-03-19
Verilog: simple immediate cover statement

kroening:
The parser already created `ID_verilog_immediate_cover` nodes for `cover(expr)`, but the typecheck and synthesis stages did not handle this ID. This adds the missing dispatch and cover-property wrapping in both stages, plus a regression test.

### Changes
- **verilog_typecheck.cpp**: Add `ID_verilog_immediate_cover` to the `convert_statement` dispatch and the kind-string mapping (so unnamed properties get `cover.N` names).
- **verilog_synthesis.cpp**: Add `ID_verilog_immediate_cover` to the `synth_statement` dispatch and wrap the condition with `sva_cover_exprt` (existential/cover semantics) in both the comment and value paths.
- **regression/verilog/SVA/immediate_cover1**: Test with a passing and failing immediate cover property.

### DIFF
--- a/src/verilog/verilog_synthesis.cpp
+++ b/src/verilog/verilog_synthesis.cpp
@@ -2403,6 +2403,7 @@ void verilog_synthesist::synth_assert_assume_cover(
       cond_for_comment = sva_assume_exprt(cond_for_comment);
     }
     else if(
+      statement.id() == ID_verilog_immediate_cover ||
       statement.id() == ID_verilog_cover_property ||
       statement.id() == ID_verilog_cover_sequence)
     {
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_synthesis.cpp
```diff
@@ -2403,6 +2403,7 @@ void verilog_synthesist::synth_assert_assume_cover(
       cond_for_comment = sva_assume_exprt(cond_for_comment);
     }
     else if(
+      statement.id() == ID_verilog_immediate_cover ||
```
tautschnig, line 2406: Just curious: why do we even need a separate ID if it seems to be handled exactly the same way as `ID_verilog_cover_property`?

kroening: There is a difference, just not enforced right now: the `cover property` statements/items need a clock, whereas the immediate variant does not.

PR #1746, opened 2026-03-22
include constrained inputs in recurrence diameter

kroening:
This adds inputs to the sum of state bits when they are constrained, either by the intial state predicate, or the next-state relation, and hence, act like state variables.

### DIFF
--- a/src/ebmc/completeness_threshold.cpp
+++ b/src/ebmc/completeness_threshold.cpp
@@ -9,6 +9,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #include "completeness_threshold.h"
 
 #include <util/arith_tools.h>
+#include <util/find_symbols.h>
 #include <util/pointer_offset_size.h>
 #include <util/std_types.h>
 
@@ -16,10 +17,13 @@ Author: Daniel Kroening, dkr@amazon.com
 #include <temporal-logic/ltl.h>
 #include <temporal-logic/nnf.h>
 #include <temporal-logic/temporal_logic.h>
+#include <trans-word-level/next_symbol.h>
 #include <verilog/sva_expr.h>
 
 #include "bmc.h"
 #include "transition_system.h"
+
+#include <unordered_set>
 
 /// An upper bound on the recurrence diameter of the state space.
 /// A purely combinational circuit (no state variables) has diameter 0.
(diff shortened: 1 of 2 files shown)


src/ebmc/completeness_threshold.cpp
```diff
@@ -47,6 +50,44 @@ recurrence_diameter(const transition_systemt &transition_system)
     total_bits += *bits;
   }
 
+  // find the variables used in the initial state constraint
+  std::unordered_set<symbol_exprt, irep_hash> init_vars;
```
tautschnig: Do we actually need to `symbol_exprt`, or would the identifier suffice?

src/ebmc/completeness_threshold.cpp
```diff
@@ -47,6 +50,44 @@ recurrence_diameter(const transition_systemt &transition_system)
...
+      }
+    });
+
+  // Inputs that are either mentioned in the initial state constraint
+  // or have a next-state constraint are not unconstraint.
+  for(auto &var : transition_system.inputs())
+  {
+    if(
+      init_vars.find(var) != init_vars.end() ||
+      next_vars.find(var) != next_vars.end())
```
tautschnig, line 68: Wouldn't it be better to just have a single (unordered) set of variables rather than having to search two of them?

src/ebmc/completeness_threshold.cpp
```diff
@@ -47,6 +50,44 @@ recurrence_diameter(const transition_systemt &transition_system)
...
 
+  // find the variables used in the initial state constraint
+  std::unordered_set<symbol_exprt, irep_hash> init_vars;
+
+  transition_system.trans_expr.init().visit_pre(
+    [&init_vars](const exprt &node)
+    {
+      if(node.id() == ID_symbol)
+        init_vars.insert(to_symbol_expr(node));
+    });
```
tautschnig: This could use `find_symbols`, but alternatively the search for `ID_symbol`/`ID_next_symbol` could also be done in a single traversal.

src/ebmc/completeness_threshold.cpp
kroening: We need the identifier and the type (to get the width); then that's almost the entire expression.

src/ebmc/completeness_threshold.cpp
kroening: I felt that this adds clarity; but can certainly merge both sets.

src/ebmc/completeness_threshold.cpp
kroening: How using `find_symbols`. Note the two traversals are done on different expressions.

tautschnig: My apologies, I had failed to spot that those are different expressions being traversed.

PR #1751, opened 2026-03-24
Sequences: weak semantics for goto and non-consecutive repetition

### LINKED ISSUE #1750: SVA [->1] (goto) operator incorrectly refuted at bound limit compared to equivalent [*0:$] sequence
Consider this minimal example:
```
module goto(input logic clk, req);

    logic ack = 1'b0;

    // This is refuted 
    assert_goto: assert property (@(posedge clk) req |=> ack[->1]);

    // This is equivalent, but proved up to bound
    assert_equivalent: assert property ( @(posedge clk) req |=> (!ack)[*0:$] ##1 ack);

endmodule
```
The goto repetition operator `b[->1]` is equivalent to `(!b)[*0:$] ##1 b` but the ebmc result is different for the two assertions:

```
goto.assert_goto] always (goto.req |=> (goto.ack [->1])): REFUTED
[goto.assert_equivalent] always (goto.req |=> ((!goto.ack [*0:$]) ##1 goto.ack)): PROVED up to bound 30

```
command issued:
ebmc -top goto   --bound 30   --systemverilog   goto.sv

ebmc version:  EBMC 5.10 (ebmc-5.9-212-g49144b0c) 64-bit  
o.s. Ubuntu 22.04

kroening:
The goto repetition (`[->n]`) and non-consecutive repetition (`[=n]`) handlers in `src/trans-word-level/sequence.cpp` ignored the weak/strong semantics parameter. Under weak semantics (the default for `assert property`), sequences that cannot complete within the bound should not be considered failures.

This caused `ack[->1]` to be incorrectly REFUTED while the semantically equivalent `(!ack)[*0:$] ##1 ack` was correctly PROVED up to bound.

The fix adds a pending match when the repetition count hasn't reached the required minimum at the end of the bound, analogous to how the cycle delay handler already treats timeframes beyond the bound.

Fixes #1750.

### DIFF
--- a/src/trans-word-level/sequence.cpp
+++ b/src/trans-word-level/sequence.cpp
@@ -376,9 +376,11 @@ sequence_matchest instantiate_sequence_rec(
     return instantiate_sequence_rec(
       repetition.lower(), semantics, t, no_timeframes);
   }
-  else if(expr.id() == ID_sva_sequence_goto_repetition) // [->...]
+  else if(
+    expr.id() == ID_sva_sequence_goto_repetition ||          // [->...]
+    expr.id() == ID_sva_sequence_non_consecutive_repetition) // [=...]
   {
-    auto &repetition = to_sva_sequence_goto_repetition_expr(expr);
+    auto &repetition = to_sva_sequence_repetition_expr(expr);
     auto &condition = repetition.op();
 
     sequence_matchest result;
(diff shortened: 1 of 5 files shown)


src/trans-word-level/sequence.cpp
```diff
@@ -431,6 +445,20 @@ sequence_matchest instantiate_sequence_rec(
...
+    if(semantics == sva_sequence_semanticst::WEAK)
+    {
+      auto min = repetition.is_range()
+                   ? numeric_cast_v<mp_integer>(repetition.from())
+                   : numeric_cast_v<mp_integer>(repetition.repetitions());
+      result.emplace_back(
+        no_timeframes - 1,
+        binary_relation_exprt{
+          matches, ID_lt, from_integer(min, matches.type())});
+    }
```
tautschnig, line 421: Except for the comments in lines 401/402 and 444, respectively, we literally have the same code twice? I believe this warrants a refactoring.

kroening: Done. The difference between `[->n]` and `[=n]` doesn't seem to be implemented. I'll investigate.

src/trans-word-level/sequence.cpp
```diff
@@ -400,35 +402,23 @@ sequence_matchest instantiate_sequence_rec(
 
       // We have a match for op[->n] if there is a match in timeframe
       // u and matches is n.
+      // We have a match for op[=n] if matches is n.
+      result.emplace_back(u, sequence_count_condition(repetition, matches));
       result.emplace_back(u, sequence_count_condition(repetition, matches));
```
tautschnig: Was duplicating this `emplace_back` intentional?

kroening: No, fixed

src/trans-word-level/sequence.cpp
tautschnig: Do you intend to fix this in this PR, or does this take a separate PR?

kroening: Will make it a separate PR.

PR #1768, opened 2026-03-27
Verilog: index-keyed array assignment patterns

kroening:
Implement type-checking for array assignment patterns with index keys and default values, per IEEE 1800-2017 section 10.9.2.


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -219,6 +219,64 @@ void verilog_typecheck_exprt::assignment_conversion(
...
+        exprt::operandst initializers(
+          numeric_cast_v<std::size_t>(array_size), nil_exprt{});
+
+        // Find the default value, if any.
+        std::optional<exprt> default_value;
+        for(auto &op : rhs.operands())
+        {
+          if(op.id() == ID_verilog_assignment_pattern_default)
+          {
+            default_value = to_unary_expr(op).op();
```
tautschnig, line 242: Shouldn't there be a check that there is at most one default value?

kroening: Done (by Kiro)

PR #1769, opened 2026-03-29
EBMC: liveness lemma engine

kroening:
Adds a heuristic engine that identifies safety properties of the form

```systemverilog
assert property (@(posedge clk) !done |=> rank < $past(rank));
```

as ranking function lemmas and uses them to prove liveness properties of the form

```systemverilog
assert property (@(posedge clk) s_eventually done);
```

The engine is added to the heuristic engine selection pipeline, between one-induction and BMC-bound-5. It normalizes each property's original expression (pre-$past instrumentation) to pattern-match the ranking function lemma structure, extracts the ranking function and done condition, then calls `is_ranking_function` to verify the inductive proof obligation.

Also promotes `engine-heuristic/ranking1` from KNOWNBUG to CORE.


src/ebmc/liveness_lemma_engine.cpp
```diff
@@ -0,0 +1,195 @@
...
+    // $past with 1 tick
+    auto ticks_opt = numeric_cast<mp_integer>(past.ticks());
+    if(!ticks_opt || *ticks_opt != 1)
+      continue;
+
+    // rank < $past(rank) — the two rank expressions must match
+    if(lt.lhs() != past.what())
+      continue;
+
+    return ranking_lemmat{strip_double_negation(candidate_done), lt.lhs()};
```
tautschnig: What about using the simplifier as a more general version of `strip_double_negation`?

kroening: Done

PR #1807, opened 2026-04-17
Verilog: multiple top-level modules

kroening:
This extends the Verilog language to support multiple top-level modules, following SystemVerilog 1800-2017 23.3.1.

### DIFF
--- a/src/verilog/verilog_ebmc_language.cpp
+++ b/src/verilog/verilog_ebmc_language.cpp
@@ -180,7 +180,7 @@ void verilog_ebmc_languaget::typecheck_module(
 
 transition_systemt verilog_ebmc_languaget::typecheck(
   const parse_treest &parse_trees,
-  irep_idt top_level_module,
+  const std::vector<irep_idt> top_level_modules,
   symbol_tablet &&symbol_table)
 {
   std::map<irep_idt, modulet> module_map;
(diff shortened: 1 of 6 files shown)


src/verilog/verilog_ebmc_language.cpp
```diff
@@ -265,29 +272,47 @@ verilog_ebmc_languaget::top_level_module(const parse_treest &parse_trees) const
...
     }
-    else
-      return given_module; // done
+
+    // now sort alphabetically
+    std::sort(
+      given_modules.begin(),
+      given_modules.end(),
+      [](const irep_idt &a, const irep_idt &b)
+      { return id2string(a) < id2string(b); });
```
tautschnig: We should probably have a single helper for this, but `symbol_table_baset::sorted_symbol_names` uses a slightly different implementation.

src/verilog/verilog_ebmc_language.cpp
```diff
@@ -309,76 +334,82 @@ verilog_ebmc_languaget::top_level_module(const parse_treest &parse_trees) const
...
+    result.push_back(module);
 
-    log.error() << messaget::eom;
-    throw ebmc_errort{}.with_exit_code(1);
-  }
+  std::sort(
+    result.begin(),
+    result.end(),
+    [](const irep_idt &a, const irep_idt &b)
+    { return id2string(a) < id2string(b); });
```
tautschnig: We should _really_ have a single common helper :-)

tautschnig (approved):

src/verilog/verilog_ebmc_language.cpp
kroening: Now in a helper.

PR #1808, opened 2026-04-17
Verilog: introduce `$unit` scope to parser

kroening:
This adds the `$unit` scope to the parser scope stack.
    
The parser to explicitly put modules and packages into the top (definitions) scope, not the current scope, which is now the `$unit` scope.

### DIFF
--- a/src/verilog/parser.y
+++ b/src/verilog/parser.y
@@ -629,7 +629,9 @@ module_identifier_with_scope:
           {
             $$ = $1;
             auto base_name = stack_expr($1).get(ID_base_name);
-            push_scope(base_name, ".", verilog_scopet::MODULE);
+            // modules go into the top scope, not into $unit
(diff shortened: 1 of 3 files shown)


tautschnig (approved):
Why does this not affect any test? Are all our tests too simple?

kroening:
The problem here is that the name resolution done by most (but not all) tools is contextual. E.g., most tools accept

```
int some_identifier;

module some_identifier;
  initial assert(some_identifier == 0);
endmodule
```

kroening:
I.e., the right thing to do is likely to switch to a recursive descent parser, and then do lookups in different name spaces depending on syntax.

tautschnig:
Does the Verilog standard not specify what the lookup rules are? Or perhaps it does and tools just ignore what the standard says?

kroening:
The clarity of the standard leaves room for improvement. See comments on test in #1818. ChatGPT says it's not valid. Claude says it's valid.

PR #1812, opened 2026-04-20
use FPA theory when generating SMT-LIB2 for Z3

kroening:
This enables the use of the FPA theory when generating SMT-LIB2.


tautschnig (commented):
1. Do all solvers that can be invoked support FPA?
2. This change also makes expressions/cast_to_real3.desc pass.

kroening:
> 1. Do all solvers that can be invoked support FPA?

We don't really know. CVC3, CVC4 and YICES support should likely be removed, given that these solvers aren't maintained. BOOLECTOR doesn't support FPA. BITWUZLA and MATHSAT support it according to their webpages, but we haven't got that tested at all.

> 2. This change also makes expressions/cast_to_real3.desc pass.

Yes, but we don't have a mechanism for "test works with SMT, but not with flattening solver". I'll first fix this in the flattening solver.

kroening:
Now limited to Z3. Note that this is already the default when using Bitwutzla or CVC 5.

tautschnig (approved):

PR #1816, opened 2026-04-21
Verilog: create `$root` module instance

kroening:
Create a $root module instance that contains an instance of the top-level module, per IEEE 1800-2017.

## Changes

- **$root module creation**: After typechecking the top-level module, a synthetic $root module is created with a `verilog_instt` that instantiates the top-level module. `verilog_synthesis` expands this, copying all symbols into $root's namespace.

- **main_symbol points to $root**: The transition system's `main_symbol` now points to `Verilog::$root`. The $root module's type is copied from the top-level module (including ports).

- **Display**: Pretty names and the expression printer strip the $root prefix since it is implicit in SystemVerilog. Internal identifiers retain $root for correct lookups.

- **Regression tests**: Updated test descriptors to reflect $root in hierarchy display, identifiers, and variable map output.

### DIFF
--- a/src/verilog/verilog_ebmc_language.cpp
+++ b/src/verilog/verilog_ebmc_language.cpp
@@ -28,6 +28,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #include "verilog_preprocessor.h"
 #include "verilog_synthesis.h"
 #include "verilog_typecheck.h"
+#include "verilog_types.h"
 
 #include <fstream>
 #include <functional>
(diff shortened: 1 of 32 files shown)


src/verilog/verilog_ebmc_language.cpp
```diff
@@ -326,14 +327,88 @@ verilog_ebmc_languaget::top_level_module(const parse_treest &parse_trees) const
...
+
+  // Create a module instance symbol for the top-level module under $root
+  symbolt instance_symbol;
+  instance_symbol.name = instance_identifier;
+  instance_symbol.base_name = top_level_module;
+  instance_symbol.pretty_name = top_level_module;
+  instance_symbol.mode = ID_Verilog;
+  instance_symbol.module = root_identifier;
+  instance_symbol.type = typet(ID_module_instance);
+  instance_symbol.value.set(ID_module, module_identifier);
```
tautschnig:
```suggestion
  symbolt instance_symbol{instance_identifier, typet{ID_module_instance}, ID_Verilog};
  instance_symbol.base_name = top_level_module;
  instance_symbol.pretty_name = top_level_module;
  instance_symbol.module = root_identifier;
  instance_symbol.value.set(ID_module, module_identifier);
```

src/verilog/verilog_ebmc_language.cpp
```diff
@@ -326,14 +327,88 @@ verilog_ebmc_languaget::top_level_module(const parse_treest &parse_trees) const
...
+  auto &top_symbol = symbol_table.lookup_ref(module_identifier);
+
+  symbolt root_symbol;
+  root_symbol.name = root_identifier;
+  root_symbol.base_name = "$root";
+  root_symbol.pretty_name = "$root";
+  root_symbol.mode = ID_Verilog;
+  root_symbol.module = root_identifier;
+  root_symbol.type = top_symbol.type;
+  root_symbol.value = verilog_module_exprt({std::move(inst)});
```
tautschnig:
```suggestion
  symbolt root_symbol{root_identifier, top_symbol.type, ID_Verilog};
  root_symbol.base_name = "$root";
  root_symbol.pretty_name = "$root";
  root_symbol.module = root_identifier;
  root_symbol.value = verilog_module_exprt({std::move(inst)});
```

src/verilog/verilog_ebmc_language.cpp
```diff
@@ -326,14 +327,88 @@ verilog_ebmc_languaget::top_level_module(const parse_treest &parse_trees) const
...
+
+  // Update port identifiers to reflect the $root hierarchy
+  for(auto &port : to_module_type(root_symbol.type).ports())
+  {
+    auto old_id = id2string(port.identifier());
+    // Replace Verilog::MODULE with Verilog::$root.MODULE
+    if(
+      old_id.compare(
+        0, module_identifier.size(), id2string(module_identifier)) == 0)
+    {
```
tautschnig, line 384:
```suggestion
    if(has_prefix(old_id, id2string(module_identifier)))
```
but then we should also introduce a `std::string module_identifier_str = id2string(module_identifier);` beforehand to avoid the repeated string-table lookup.

src/verilog/verilog_ebmc_language.cpp
```diff
@@ -326,14 +327,88 @@ verilog_ebmc_languaget::top_level_module(const parse_treest &parse_trees) const
...
+      port.identifier(
+        id2string(instance_identifier) +
+        old_id.substr(module_identifier.size()));
+    }
+  }
+
+  symbol_table.add(root_symbol);
```
tautschnig: We should assert that this was successfully added.

src/verilog/verilog_synthesis.cpp
```diff
@@ -1729,7 +1729,12 @@ void verilog_synthesist::expand_module_instance(
       std::string full_identifier =
         id2string(instance.identifier()) + identifier_without_module;
 
-      new_symbol.pretty_name=strip_verilog_prefix(full_identifier);
+      // We special-case the pretty name for identifiers under $root:
+      // it is customary to use the name of the module only as root
+      // of the hierarchical identifier.
+      new_symbol.pretty_name = module == verilog_module_symbol("$root")
```
tautschnig: We should have `$root` as either a string constant or a macro, or maybe just `ID_dollar_root` rather than repeating the string literal.

tautschnig: Hmm, where has this been done? I can still see several appearances of `"$root"`.

kroening: Added `ID_dollar_root`

kroening: And now a `const` global variable.

src/verilog/verilog_typecheck_base.cpp
```diff
@@ -17,6 +17,8 @@ Author: Daniel Kroening, kroening@kroening.com
 #include "verilog_bits.h"
 #include "verilog_types.h"
 
+const irep_idt verilog_root_module_name = "$root";
```
tautschnig: I believe this works nowadays, but we will need to revisit the comments at `dstringt(const char *s)` and `dstringt(const std::string &s)` (which say: "this one is not safe for static objects").

tautschnig (approved):

src/verilog/verilog_typecheck_base.cpp
kroening: Will turn into function for now.

PR #1822, opened 2026-04-23
test to determine binding of `else` vs. `if` action blocks

kroening:
This test determines whether an `else` binds to the most recent assertion as part of an `action_block`, or to the most recent `if`, as part of an `conditional_statement`.

### DIFF
--- a/regression/verilog/SVA/immediate4.sv
+++ b/regression/verilog/SVA/immediate4.sv
@@ -6,13 +6,11 @@ module main(input clk);
 
   always @(posedge clk) begin : blk
 
-    // It is unclear whether this should parse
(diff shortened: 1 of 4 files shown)


kroening:
Claude wins vs. ChatGPT!

tautschnig (commented):
Could we please have data for the other Verilog processing tools like Icarus etc to see what they do?

kroening:
> Could we please have data for the other Verilog processing tools like Icarus etc to see what they do?

All in `immediate5.sv‎`! There is a clear majority for "binds to assert".

tautschnig (approved):

PR #1831, opened 2026-04-25
AIGER: complete the AIGER front-end

kroening:
## Summary

Complete the AIGER front-end to read AIGER files directly, removing the need for the `aigtosmv` conversion to SMV. Both the binary format (`.aig`) and the ASCII format (`.aag`) are supported.

## Changes

- **`src/aiger/aiger_ebmc_language.{h,cpp}`**: New `ebmc_languaget` implementation that parses AIGER files using the aiger C library and builds a transition system with inputs, latches, AND gates, outputs, and properties.
- **`src/aiger/aiger_language.{h,cpp}`**: Updated the CBMC-style `languaget` interface with working implementations. Registers both `.aig` and `.aag` extensions.
- **`src/aiger/Makefile`**: Compiles `aiger.c` from `src/ic3/aiger/` into the aiger library.
- **`src/ebmc/build_transition_system.cpp`**: Added `.aig` and `.aag` extension dispatch.
- **`benchmarking/hwmcc08.sh`**: Simplified to use `.aig` files directly (no more `aigtosmv` dependency).
- **`regression/ebmc/aiger/`**: Six regression tests covering both format versions and property types, with trace oracles.

## Format support

- **Binary AIGER** (`.aig`): The compact binary encoding used by HWMCC benchmarks
- **ASCII AIGER** (`.aag`): The human-readable text format

Both formats are auto-detected by the underlying aiger library.

## Property support

- **Safety (AIGER 1.0)**: Outputs treated as bad-state signals → `G(!output)`
- **Safety (AIGER 1.9+)**: Explicit bad signals → `G(!bad)`
- **Liveness**: Justice constraints → `GF(l₁ ∧ l₂ ∧ ...)`, fairness → `GF(lit)`

## Performance: AIGER vs SMV flow on HWMCC08

Compared direct AIGER reading vs the previous flow (aigtosmv → SMV → ebmc) on 433 HWMCC08 benchmarks with known results:

| Metric | AIGER | SMV |
|--------|-------|-----|
| Total time | 75.4s | 100.7s |
| Faster on | 426/433 | 7/433 |
| **Speedup** | **1.34×** | — |
| Result mismatches | **0** | — |

The speedup comes from eliminating the aigtosmv conversion overhead and the SMV parsing/typechecking step. All 433 benchmarks produce identical verdicts.

## Testing

- 6 regression tests: safe/unsafe × AIGER 1.0/1.9, liveness pass/fail
- All existing regression tests pass (no regressions)
- All 585 HWMCC08 benchmarks run without errors


src/aiger/aiger_ebmc_language.cpp
```diff
@@ -0,0 +1,363 @@
...
+    irep_idt id = id2string(module_id) + "::var::" + name;
+
+    symbolt symbol{id, bool_typet(), mode};
+    symbol.base_name = name;
+    symbol.pretty_name = name;
+    symbol.module = module_id;
+    symbol.is_input = true;
+    symbol.is_state_var = false;
+
+    result.symbol_table.add(symbol);
```
tautschnig: Here and elsewhere: shouldn't we be checking the return value to make sure the symbol genuinely was added?

src/aiger/aiger_ebmc_language.cpp
```diff
@@ -0,0 +1,363 @@
...
+    symbol.is_state_var = true;
+
+    result.symbol_table.add(symbol);
+    var_map[model.latches[i].lit >> 1] = id;
+  }
+
+  // AND gates
+  for(std::size_t i = 0; i < model.num_ands; i++)
+  {
+    unsigned lhs_var = model.ands[i].lhs >> 1;
```
tautschnig, line 161: Just for consistency reasons: this appears to be the only case where we use a named temporary for this expression.

src/aiger/aiger_ebmc_language.cpp
```diff
@@ -0,0 +1,363 @@
...
+  for(std::size_t i = 0; i < model.num_latches; i++)
+  {
+    irep_idt latch_id = var_map[model.latches[i].lit >> 1];
+    symbol_exprt latch_sym(latch_id, bool_typet());
+
+    unsigned reset = model.latches[i].reset;
+    if(reset == 0)
+      init_exprs.push_back(equal_exprt(latch_sym, false_exprt()));
+    else if(reset == 1)
+      init_exprs.push_back(equal_exprt(latch_sym, true_exprt()));
```
tautschnig: Should there be a guard that no other value is seen?

tautschnig (approved):

src/aiger/aiger_ebmc_language.cpp
kroening: done

src/aiger/aiger_ebmc_language.cpp
kroening: added

PR #1832, opened 2026-04-26
engine heuristic: copy result only

kroening:
Instead of copying the entire property data structure, the engine heuristic now only copies the fields that store the result.

This enables engines to make local changes to the other fields.

### DIFF
--- a/src/ebmc/engine_heuristic.cpp
+++ b/src/ebmc/engine_heuristic.cpp
@@ -92,6 +92,21 @@ const enginet engines[] = {
   {bmc_bound_5_engine, "BMC with bound 5"},
 };
 
+void copy_results_to(
+  ebmc_propertiest::propertiest &src,
+  ebmc_propertiest::propertiest &dest)
+{
+  PRECONDITION(src.size() == dest.size());
+
+  // This could use std::views::zip with C++23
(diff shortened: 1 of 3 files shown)


src/ebmc/engine_heuristic.cpp
```diff
@@ -92,6 +92,25 @@ const enginet engines[] = {
...
+  for(auto src_it = src.begin(), dest_it = dest.begin(); src_it != src.end();
+      src_it++, dest_it++)
+  {
+    // copy the result fields of the property
+    dest_it->status = src_it->status;
+    dest_it->bound = src_it->bound;
+    dest_it->witness_trace = src_it->witness_trace;
+    dest_it->failure_reason = src_it->failure_reason;
+    dest_it->proof_via = src_it->proof_via;
+  }
```
tautschnig, line 107: As this is an `ebmc_propertiest::propertiest` implementation detail, wouldn't it be better to make this a member function of `ebmc_propertiest::propertiest` (something like `ebmc_propertiest::propertiest::copy_results_from(const propertiest &src)`)?

tautschnig (approved):

src/ebmc/engine_heuristic.cpp
kroening: Making it a method

PR #1836, opened 2026-04-28
BMC: semantics of pending matches

kroening:
## Summary

Adds an `allow_pending_matches` parameter to the BMC property obligation
pipeline, controlling whether pending sequence matches (those that cannot
complete within the verification bound) are kept or dropped.

## Semantic change

In BMC we only search for counterexamples on finite paths. A pending match
is inconclusive — it neither proves nor disproves the property.

- **Universal properties (assert):** `allow_pending_matches=true`. Pending
  matches are kept as vacuously true. No counterexample was found at the
  boundary, so BMC correctly reports "PROVED up to bound".
- **Existential properties (cover):** `allow_pending_matches=false`. Pending
  matches are dropped — an incomplete match is not a witness.

Previously, the strong/weak sequence distinction controlled this: strong
sequences dropped pending matches, weak sequences kept them. This was
incorrect for BMC — the strong/weak distinction is about whether a sequence
*requires* a match to exist, which is a separate concern from how BMC handles
the finite-path boundary. The `sva_sequence_semantics` helper is removed.

Implication and followed-by antecedents continue to unconditionally drop
pending LHS matches, since a pending antecedent should not trigger the
consequent.

## Effect on tests

Properties like `assert property (strong(##[0:9] x==5))` with a bound too
low for the match to complete now report "PROVED up to bound" instead of
"REFUTED". The old REFUTED was unsound — BMC was conflating "match cannot
complete within bound" with "property is refuted".

## What was tested

- All verilog regression tests pass
- All ebmc regression tests pass

### DIFF
--- a/regression/verilog/SVA/sequence2.bmc.buechi.desc
+++ b/regression/verilog/SVA/sequence2.bmc.buechi.desc
@@ -1,8 +1,9 @@
 CORE buechi
 sequence2.sv
 --buechi --bound 10
-^\[main\.p0] weak\(##\[0:\$\] main\.x == 10\): PROVED up to bound 10$
-^\[main\.p1] strong\(##\[0:\$\] main\.x == 10\): REFUTED$
+^\[main\.p0\] weak\(##\[0:\$\] main\.x == 10\): PROVED up to bound 10$
+^\[main\.p1\] strong\(##\[0:\$\] main\.x == 10\): REFUTED$
+^\[main\.c0\] cover strong\(##\[0:\$\] main\.x == 10\): REFUTED up to bound 10$
 ^EXIT=10$
 ^SIGNAL=0$
 --
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,5 +1,6 @@
 # EBMC 6.0
 
+* BMC: strong sequences that exceed the bound are no longer reported as REFUTED
 * VCD traces now written into separate files per property
 * SMV: removed legacy keywords EXTERN and switch
 * SMV: use NuSMV's operator precedence, as opposed to CMU SMV's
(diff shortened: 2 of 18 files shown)


regression/verilog/SVA/named_sequence1.desc
```diff
@@ -1,7 +1,7 @@
 CORE
 named_sequence1.sv
 
-^\[main\.assert\.1\] always main\.x_is_ten: PROVED \(transition property\)$
+^\[main\.assert\.1\] always main\.x_is_ten: PROVED up to bound 5$
```
tautschnig: Shouldn't "transition property" proofs remain unbounded?

regression/verilog/SVA/sequence2.bmc.buechi.desc
```diff
@@ -1,8 +1,9 @@
 CORE buechi
 sequence2.sv
 --buechi --bound 10
-^\[main\.p0] weak\(##\[0:\$\] main\.x == 10\): PROVED up to bound 10$
-^\[main\.p1] strong\(##\[0:\$\] main\.x == 10\): REFUTED$
+^\[main\.p0\] weak\(##\[0:\$\] main\.x == 10\): PROVED up to bound 10$
+^\[main\.p1\] strong\(##\[0:\$\] main\.x == 10\): REFUTED$
```
tautschnig, line 5: The side-by-side diff is suggesting something changed here, but I don't quite see what the diff is.

kroening: Missing `\` before the first `]`, on both lines.

regression/verilog/SVA/named_sequence1.desc
kroening: I'll need to investigate why the transition property engine no longer proves this.

kroening: Now fixed

tautschnig (approved):

PR #1839, opened 2026-04-30
bump cbmc dependency to release 6.10.0

kroening:


regression/smv/range-type/range_type11.desc
```diff
@@ -1,8 +1,9 @@
...
+KNOWNBUG broken-smt-backend
 range_type11.smv
 
 ^EXIT=0$
 ^SIGNAL=0$
 --
 --
-The range constraint is missing.
+The range constraint is missing for the flattening backend.
+Works with SMT.
```
tautschnig, line 9: I'm a bit confused - it's tagged `broken-smt-backend`?

tautschnig (approved):

regression/smv/range-type/range_type11.desc
kroening: Yes, but you need to negate that in your head when it's a KNOWNBUG.

PR #1841, opened 2026-05-01
bad pr please ignore

matx-adenko:
wrong target, my bad


kroening:
> wrong target, my bad

We indeed should get `hw-cbmc` working again.

PR #1842, opened 2026-05-04
Verilog: grammar for classes

kroening:
This adds grammar for 1800 2017 8.3 classes, including properties with qualifiers, methods, constructors, prototypes, constraints, virtual classes, extends/implements, and 8.26 interface classes. Class items are stored in the parse tree.

The optional const/lifetime prefixes of data_declaration are factored out to keep the grammar LALR(1) conflict-free.


tautschnig:
Should there be a test?  Also, the comment in the parser still says "classes are yet to be implemented".

kroening:
Will finish, will then have tests.

tautschnig (approved):

PR #1843, opened 2026-05-04
SMV netlists: use pretty name, when available

kroening:
When printing SMV netlists use the pretty name of the symbol Instead of using the full identifier, when available.

### DIFF
--- a/src/trans-netlist/smv_netlist.cpp
+++ b/src/trans-netlist/smv_netlist.cpp
@@ -129,14 +129,60 @@ void print_smv(
   out << expr2smv_netlistt{ns, netlist, variable_names}.convert(expr);
 }
 
-void smv_netlist(const netlistt &netlist, std::ostream &out)
+std::map<literalt::var_not, std::string> smv_names(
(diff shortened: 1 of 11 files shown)


src/trans-netlist/smv_netlist.cpp
```diff
@@ -129,14 +129,62 @@ void print_smv(
...
-void smv_netlist(const netlistt &netlist, std::ostream &out)
+std::map<literalt::var_not, std::string> smv_names(
+  const std::vector<var_mapt::mapt::const_iterator> &sorted_identifiers,
+  const namespacet &ns)
+{
+  std::map<literalt::var_not, std::string> smv_names;
+
+  auto map_name = [&smv_names, &ns](var_mapt::mapt::const_iterator var_it)
+  {
+    {
```
tautschnig, line 142: This block seems unnecessary.

tautschnig (approved):
Looks like more tests need removal of the `Verilog` prefix from .desc files.

src/trans-netlist/smv_netlist.cpp
kroening: fixed

PR #1847, opened 2026-05-09
SMV netlist: escape SMV keywords in identifiers

kroening:
When a Verilog signal name coincides with an SMV keyword (e.g., MODULE), the generated SMV netlist must escape it with a backslash prefix to produce valid SMV syntax.

This adds a post-processing step to `id2smv()` that checks each dot-separated component against `is_smv_keyword()` and prefixes matches with `\`, producing e.g., `main.\MODULE`.

### DIFF
--- a/src/trans-netlist/smv_netlist.cpp
+++ b/src/trans-netlist/smv_netlist.cpp
@@ -10,8 +10,10 @@ Author: Daniel Kroening, dkr@amazon.com
 #include <smvlang/expr2smv_class.h>
 
 #include <util/namespace.h>
+#include <util/string_utils.h>
 #include <util/symbol_table.h>
 
+#include <smvlang/smv_keywords.h>
 #include <solvers/prop/literal_expr.h>
 #include <temporal-logic/temporal_logic.h>
 #include <verilog/sva_expr.h>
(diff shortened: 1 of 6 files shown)


src/trans-netlist/smv_netlist.cpp
```diff
@@ -44,7 +45,23 @@ std::string id2smv(const irep_idt &id)
...
   }
 
-  return result;
+  // Escape any dot-separated components that are SMV keywords
+  std::string escaped;
+  std::size_t pos = 0;
+  while(pos < result.size())
+  {
+    auto dot = result.find('.', pos);
+    auto component = result.substr(pos, dot - pos);
```
tautschnig: What if `dot` is std::string::npos (aka -1)?

kroening: It does do the right thing (the `len` argument of `substr` is allowed to be larger than the number of remaining characters after `pos`), but I'll refactor to make this clear.

kroening: now uses `split_string`

tautschnig (approved):

PR #1848, opened 2026-05-10
RFC: pyebmc Python bindings design

kroening:
This is a Request for Comments for the design of `pyebmc`, a Python bindings package for EBMC.

## Summary

- EDA-familiar API (`read_verilog`, `elaborate`, `prove`) on a `Design` session object
- CBMC IR types (`Expr`, `Type`) in a separate `pyebmc.cprover` namespace (to eventually become its own package)
- Property iteration with full expression tree access
- Constraint API for per-time-frame assumptions
- pybind11 binding layer

## Feedback requested

- API naming and conventions
- Namespace separation (`pyebmc` vs `pyebmc.cprover`)
- Engine/solver parameter design
- Expression construction API (future)

### DIFF
--- a/doc/pyebmc-design.md
+++ b/doc/pyebmc-design.md
@@ -1,0 +1,1037 @@
+# pyebmc — Python Bindings for EBMC
+
+## Design Document (RFC)
+
+**Status:** Request for Comments  
+**Date:** 2026-05-10
+
+### Goals
+
+1. Provide a Pythonic API for EBMC that feels natural to EDA engineers
+   accustomed to Tcl-based tool scripting (Cadence, Synopsys, Mentor).
+2. Follow standard Python conventions (PEP 8, context managers, iterators,
+   properties, exceptions).
+3. Mirror the EDA Tcl idiom of imperative "command" calls that configure and
+   query a persistent session/database, while wrapping them in proper Python
+   objects.
+
+### Design Principles
+
+| EDA Tcl Convention | pyebmc Python Equivalent |
+|-|-|
+| Global tool database / session | `ebmc.Design` object (explicit session) |
+| `read_verilog`, `read_hdl` commands | `design.read_verilog(...)` methods |
+| `set_top`, `elaborate` | `design.set_top(...)`, `design.elaborate()` |
+| `check_property`, `prove` | `design.check_property(...)` |
+| `report_*` commands | `design.report_properties()`, iteration |
+| `-option value` flags | Keyword arguments |
+| Return status codes | Exceptions + result enums |
+| `get_*` / `set_*` attribute queries | Python `@property` accessors |
+
+---
+
+## Package Structure
+
+```
+pyebmc/
+├── __init__.py          # Public API re-exports (Design, Property, etc.)
+├── design.py            # Design class (the session object)
+├── property.py          # Property / PropertyResult data classes
+├── trace.py             # Trace / Waveform access
+├── module.py            # Module introspection
+├── exceptions.py        # Exception hierarchy
+├── cprover/             # CBMC-derived IR types (separate namespace)
+│   ├── __init__.py      # Re-exports Expr, Type, SourceLocation
+│   ├── expr.py          # Expr class and subclass registry
+│   └── type.py          # Type class and subclass registry
+├── _engine.py           # Low-level C++ binding (pybind11)
+└── py.typed             # PEP 561 marker
+```
+
+The `pyebmc.cprover` subpackage contains types originating from the CBMC
+intermediate representation (`exprt`, `typet`, `source_locationt`). These
+are general-purpose IR nodes not specific to EBMC. The top-level `pyebmc`
+namespace contains EBMC-specific objects (`Design`, `Property`, `Trace`,
+etc.).
+
+> **Note:** The `pyebmc.cprover` subpackage is intended to eventually move
+> into its own standalone package (e.g. `pycprover`) that can be shared
+> across CBMC-based tools. For now it ships inside `pyebmc` to simplify
+> initial development and distribution.
+
+---
+
+## Core API
+
+### Session: `Design`
+
+The central object, analogous to a Tcl tool session. All state lives here.
+
+```python
+import pyebmc
+
+# Create a session (like opening a tool shell)
+design = pyebmc.Design()
+
+# --- Reading sources (like read_verilog / read_file) ---
+design.read_verilog("alu.sv")
+design.read_verilog("top.sv", systemverilog=True)
+design.read_smv("model.smv")
+
+# --- Elaboration (like elaborate / set_top) ---
+design.set_top("top")
+design.elaborate()
+
+# --- Accessing the design hierarchy ---
+for mod in design.modules:
+    print(mod.name, mod.file, mod.line)
+
+# --- Properties ---
+for prop in design.properties:
+    print(prop.name, prop.expression, prop.status)
+```
+
+#### Constructor
+
+```python
+class Design:
+    def __init__(
+        self,
+        *,
+        verbosity: int = 2,
+        solver: str = "minisat",   # "minisat", "cadical", "z3", "bitwuzla"
+    ) -> None: ...
+```
+
+#### Source Reading Commands
+
+These mirror the Tcl `read_*` family. The Tcl `set_include_path` and
+`set_define` session commands are handled as keyword arguments to
+`read_verilog` instead, keeping configuration explicit and per-call:
+
+```python
+def read_verilog(
+    self,
+    path: str | Path,
+    *,
+    systemverilog: bool = False,
+    defines: dict[str, str] | None = None,
+    include_dirs: list[str | Path] | None = None,
+) -> None: ...
+
+def read_verilog_string(
+    self,
+    source: str,
+    *,
+    systemverilog: bool = False,
+    filename: str = "<string>",
+    defines: dict[str, str] | None = None,
+) -> None:
+    """Read Verilog/SystemVerilog source from a Python string."""
+    ...
+
+def read_smv(self, path: str | Path) -> None: ...
+```
+
+#### Elaboration Commands
+
+```python
+def set_top(self, module_name: str) -> None: ...
+
+def elaborate(self) -> None:
+    """Elaborate the design. Raises ElaborationError on failure."""
+    ...
+
+def set_reset(
+    self,
+    signal: str,
+    *,
+    active_high: bool = True,
+    cycles: int = 1,
+) -> None:
+    """Define the reset sequence for verification.
+    Mirrors the Tcl set_reset_sequence idiom."""
+    ...
+```
+
+#### Verification Commands
+
+The core "prove" / "check" commands, matching the EDA idiom of running
+analysis on the current design state:
+
+```python
+def check_property(
+    self,
+    *,
+    bound: int = 10,
+    property: str | None = None,   # specific property identifier
+    engine: str = "bmc",           # "bmc", "k-induction", "k-induction-step", "ic3", "bdd", "buechi"
+    solver: str | None = None,     # override session solver
+) -> "ProveResult": ...
+```
+
+#### Constraints
+
+```python
+def add_constraint(
+    self,
+    expression: str,
+    *,
+    step: int | None = None,
+) -> None:
+    """Add a constraint to the verification.
+    If step is given, the constraint applies only at that time frame.
+    Otherwise it applies at all time frames."""
+    ...
+```
+
+#### Property Management
+
+```python
+def add_property(
+    self,
+    name: str,
+    expression: str,
+) -> None:
+    """Add a property from a script (SVA expression string)."""
+    ...
+
+def add_property_expr(
+    self,
+    name: str,
+    expression: "Expr",
+) -> None:
+    """Add a property from a programmatically constructed Expr tree."""
+    ...
+
+def disable_property(self, name: str) -> None:
+    """Disable a property so it is excluded from verification."""
+    ...
+
+def enable_property(self, name: str) -> None:
+    """Re-enable a previously disabled property."""
+    ...
+```
+
+#### Trace / Waveform Access
+
+```python
+def get_trace(self, property: str) -> "Trace | None":
+    """Get counterexample trace for a refuted property."""
+    ...
+
+def random_traces(
+    self,
+    *,
+    count: int = 1,
+    steps: int = 10,
+    seed: int | None = None,
+) -> list["Trace"]: ...
+
+def write_vcd(self, path: str | Path, property: str) -> None:
+    """Write counterexample as VCD waveform file."""
+    ...
+```
+
+#### Reporting Commands
+
+```python
+def report_properties(self, *, format: str = "text") -> str:
+    """Return formatted property report. format: 'text', 'json', 'xml'."""
+    ...
+
+def report_modules(self, *, format: str = "text") -> str: ...
+```
+
+#### State Query Properties
+
+```python
+@property
+def modules(self) -> list["Module"]: ...
+
+@property
+def properties(self) -> list["Property"]: ...
+
+@property
+def state_variables(self) -> list[str]: ...
+
+@property
+def inputs(self) -> list[str]: ...
+```
+
+---
+
+### Data Classes
+
+#### `Module`
+
+```python
+@dataclass(frozen=True)
+class Module:
+    name: str
+    identifier: str
+    mode: str              # "Verilog", "SMV", etc.
+    file: str | None
+    line: int | None
+```
+
+#### `Property`
+
+```python
+from pyebmc.cprover import Expr
+
+@dataclass
+class Property:
+    name: str
+    identifier: str
+    expr: Expr             # structured expression tree (see cprover.Expr)
+    status: PropertyStatus
+    bound: int | None
+    failure_reason: str | None
+    proof_method: str | None
+
+    @property
+    def expression(self) -> str:
+        """Human-readable rendering of the property expression."""
+        return str(self.expr)
+
+    @property
+    def proved(self) -> bool: ...
+
+    @property
+    def refuted(self) -> bool: ...
+
+    @property
+    def has_trace(self) -> bool: ...
+```
+
+#### `PropertyStatus` (Enum)
+
+```python
+class PropertyStatus(enum.Enum):
+    UNKNOWN = "unknown"
+    PROVED = "proved"
+    PROVED_WITH_BOUND = "proved_with_bound"
+    REFUTED = "refuted"
+    REFUTED_WITH_BOUND = "refuted_with_bound"
+    DISABLED = "disabled"
+    ASSUMED = "assumed"
+    UNSUPPORTED = "unsupported"
+    DROPPED = "dropped"
+    FAILURE = "failure"
+    INCONCLUSIVE = "inconclusive"
+```
+
+#### `ProveResult`
+
+```python
+@dataclass(frozen=True)
+class ProveResult:
+    properties: list[Property]
+
+    @property
+    def all_proved(self) -> bool: ...
+
+    @property
+    def has_counterexample(self) -> bool: ...
+
+    def __bool__(self) -> bool:
+        """True if all properties proved (enables `if design.check_property(): ...`)."""
+        return self.all_proved
+```
+
+#### `Trace`
+
+```python
+@dataclass(frozen=True)
+class Trace:
+    steps: int
+    signals: list[str]
+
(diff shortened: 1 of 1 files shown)


jimgrundy (approved):
* The reset sequence command might be a little too simple for more complex designs. Well, actually, I can't tell maybe the "*" on line 148 covers a lot. The other tools let you give a file that supplies a sequence of values to be pumped in over time to reset a device. I've had caches that took over a thousand cycles to bring to the initial state. That said, that mechanism always sucked. I would rather have been able to define my own reset state machine that I could attach to the design under verification for the reset period (after which it would detach letting the inputs it drove float (constrained by the environment). 

* Among the property management APIs I'm missing a command that would convert an assert to an assume and visa versa. These are really handy for compositional verification and for sharing properties between FV and DV. I couldn't live without them.

* The "get_trace" doc string says it is to get the trace of a refuted property. Is there also a notion of a cover property and this also works for that? It doesn't look that way from the PropertyStatus type, but I find those really important for checking how far I am getting into a design with bounded model checking and convincing myself that proofs aren't vacuous. 

* Is this interface rich enough to implement CAG? We built CAG for another tool using it's Tcl interface. We found it much more productive.   Maybe not yet, as it involved manipulating the net-list and I see that is a "to do" for the API.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+are general-purpose IR nodes not specific to EBMC. The top-level `pyebmc`
+namespace contains EBMC-specific objects (`Design`, `Property`, `Trace`,
+etc.).
+
+> **Note:** The `pyebmc.cprover` subpackage is intended to eventually move
+> into its own standalone package (e.g. `pycprover`) that can be shared
+> across CBMC-based tools. For now it ships inside `pyebmc` to simplify
+> initial development and distribution.
+
+---
```
tautschnig, line 62: Namespace migration plan: if `pyebmc.cprover` is going to move to a standalone `pycprover` package, please commit to shipping both import paths from day 1 (with `pyebmc.cprover` being a thin re-export of `pycprover`). Otherwise every user will have a painful `from pyebmc.cprover import …` → `from pycprover import …` migration when the split happens, and that window lasts for however long it takes the ecosystem to catch up. Fixing the preferred import path now costs nothing and is the one piece of the RFC that is most expensive to change later.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+
+```python
+class Design:
+    def __init__(
+        self,
+        *,
+        verbosity: int = 2,
+        solver: str = "minisat",   # "minisat", "cadical", "z3", "bitwuzla"
+    ) -> None: ...
+```
```
tautschnig, line 104: Two things on the `Design` constructor:

1. `solver: str` should be `Literal["minisat", "cadical", "z3", "bitwuzla"]` (or an enum). A typo silently falls back to the default string otherwise, and IDEs can't autocomplete.
2. Nothing here covers cleanup. The RFC lists "context managers" as a design principle but `Design` has no `__enter__`/`__exit__` and no `close()`. Long-running Python processes that create and discard sessions will leak solver state unless destruction is guaranteed. Please add both an explicit `close()` and the `__enter__`/`__exit__` pair.

Also worth mentioning: can there be multiple `Design` instances alive at once? CBMC has historically carried global state (`messaget`, `dstringt` tables) that makes concurrency tricky. If only one session is supported per process, say so explicitly; if not, call out the threading guarantees.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+`read_verilog` instead, keeping configuration explicit and per-call:
+
+```python
+def read_verilog(
+    self,
+    path: str | Path,
+    *,
+    systemverilog: bool = False,
+    defines: dict[str, str] | None = None,
+    include_dirs: list[str | Path] | None = None,
```
tautschnig, line 119: `include_dirs` and `defines` are per-file here, which matches Tcl's `read_verilog -define` / `-incdir` but not the more common Tcl pattern of `set_include_path` / `set_define` as session state. In practice users with large designs set these once globally. Suggest adding them as `Design.__init__` keyword args (session-wide defaults) that can still be overridden per-call, instead of forcing the user to pass the same `include_dirs=[...]` argument to every `read_verilog` call.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+    *,
+    systemverilog: bool = False,
+    filename: str = "<string>",
+    defines: dict[str, str] | None = None,
+) -> None:
+    """Read Verilog/SystemVerilog source from a Python string."""
+    ...
+
+def read_smv(self, path: str | Path) -> None: ...
+```
```
tautschnig, line 134: For consistency with `read_verilog_string`, consider adding `read_smv_string` as well. Users who programmatically construct SMV snippets (test harnesses, benchmarks) otherwise have to round-trip through a temp file.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+    self,
+    signal: str,
+    *,
+    active_high: bool = True,
+    cycles: int = 1,
+) -> None:
+    """Define the reset sequence for verification.
+    Mirrors the Tcl set_reset_sequence idiom."""
+    ...
+```
```
tautschnig, line 155: As follow-up to @jimgrundy's comment: the reset API is too shallow for non-trivial designs. A single signal + polarity + cycle count doesn't cover the real-world case of caches / DMA engines that need hundreds to thousands of cycles of coordinated input sequences before reaching the initial verification state.

Two directions worth considering:

1. Accept a `Trace`-like object (or a list of dicts `{signal: value, ...}`) describing the reset stimulus per cycle. This is the "input file" pattern Jim alluded to.
2. Allow `set_reset(module=<SVModule>)` where the user provides a synthesizable SV module to drive reset-phase inputs, which is then composed with the DUT and detached after reset.

Even if the first release only implements option 1, the API shape should leave room for option 2 so users aren't forced to rewrite code later. Right now `set_reset(signal=..., cycles=...)` paints us into the simple-case corner.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+
+```python
+def check_property(
+    self,
+    *,
+    bound: int = 10,
+    property: str | None = None,   # specific property identifier
+    engine: str = "bmc",           # "bmc", "k-induction", "k-induction-step", "ic3", "bdd", "buechi"
+    solver: str | None = None,     # override session solver
+) -> "ProveResult": ...
```
tautschnig, line 170:
```suggestion
    engine: Engine = Engine.BMC,
    solver: Solver | None = None,
) -> "ProveResult": ...
```

…with `Engine` and `Solver` as `enum.Enum` (or `typing.Literal`) defined next to `PropertyStatus`. Stringly-typed knobs like `"k-induction-step"` are fragile — IDE completion doesn't help users, and a typo silently selects the default engine. Same concern applies to `Design.__init__(solver=)` and `report_properties(format=)`.

Also: why is `property` a `str` here but a `Property` object is returned by iteration? The API would be more ergonomic if `check_property(property=...)` accepted `str | Property`, and the same rule applied to `get_trace`, `write_vcd`, `disable_property`, `enable_property`. Pick one rule and apply it consistently.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+    self,
+    expression: str,
+    *,
+    step: int | None = None,
+) -> None:
+    """Add a constraint to the verification.
+    If step is given, the constraint applies only at that time frame.
+    Otherwise it applies at all time frames."""
+    ...
+```
```
tautschnig, line 186: `add_constraint` takes an `expression: str`. What language is this parsed as? SVA? Verilog? A subset of either? This needs to be specified in the docstring and ideally matched to whatever `add_property(..., expression: str)` accepts on line 196, because right now the two string-based entry points are silently different dialects from the user's point of view.

Once `Expr`-level construction is real, please also add an `add_constraint_expr(expression: Expr, *, step=None)` variant so the constraint/property APIs are symmetric.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+    ...
+
+def disable_property(self, name: str) -> None:
+    """Disable a property so it is excluded from verification."""
+    ...
+
+def enable_property(self, name: str) -> None:
+    """Re-enable a previously disabled property."""
+    ...
+```
```
tautschnig, line 214: For @jimgrundy's prior comment that there is no API to convert an `assert` into an `assume` (or vice versa). This is essential for compositional verification (assume-guarantee) and for sharing properties between formal and dynamic verification flows. `PropertyStatus.ASSUMED` exists in the status enum, but there's no way to transition a property into that state from Python.

Suggested addition:

```python
def set_property_role(
    self,
    property: str | Property,
    role: PropertyRole,  # ASSERT | ASSUME | COVER
) -> None: ...
```

This also subsumes `disable_property` (role=`DISABLED`) if you want, or leave those as separate verbs — but the assert↔assume toggle is a hard requirement for real-world compositional flows.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+
+#### Trace / Waveform Access
+
+```python
+def get_trace(self, property: str) -> "Trace | None":
+    """Get counterexample trace for a refuted property."""
+    ...
```
tautschnig, line 221: As @jimgrundy said: the docstring is wrong for cover properties. In EBMC, a *successful* cover property produces a witness trace — `has_witness_trace()` returns true when `is_refuted()` *or* when a cover goal is hit. Restricting `get_trace` to refuted-only properties hides that use case, which is genuinely important for sanity-checking that the BMC unwinding is deep enough to exercise the interesting parts of the design.

Minimally, change the docstring to "Get counterexample or cover witness trace, if one is available." Better: expose `Property.has_trace` as the predicate users should check, and let `get_trace()` return a trace whenever `has_witness_trace()` holds on the underlying property.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+    *,
+    count: int = 1,
+    steps: int = 10,
+    seed: int | None = None,
+) -> list["Trace"]: ...
+
+def write_vcd(self, path: str | Path, property: str) -> None:
+    """Write counterexample as VCD waveform file."""
+    ...
+```
```
tautschnig, line 234: Redundant with `design.get_trace(p).to_vcd()` + a plain file write. I'd drop `write_vcd` from `Design` entirely and document the one-liner instead — it keeps the trace-handling surface in one place (on `Trace`), and makes it obvious what happens when the property has no trace (user decides, instead of `write_vcd` silently producing nothing / raising).

If we keep both, the behavior when the property has no counterexample needs to be specified (raise? return bool? write an empty file?).

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+
+    @property
+    def proved(self) -> bool: ...
+
+    @property
+    def refuted(self) -> bool: ...
+
+    @property
+    def has_trace(self) -> bool: ...
+```
```
tautschnig, line 306: `name` vs `identifier` — what's the difference, and which do users pass where? Example 1 on line 414-415 has the user grabbing `prop.identifier` to pass to `get_trace` / `write_vcd` but `prop.name` to build the filename. That's an easy way to get it wrong.

Either:
1. Collapse the two into one (probably `identifier`, keeping `name` as an `@property` alias), or
2. Spell out clearly: "`name` is the user-facing label (e.g. `p1`), `identifier` is the fully-qualified internal handle (e.g. `main.p1`); the Python API accepts either everywhere a property is referenced."

Option 2 is less disruptive. Either way, make it explicit in the docstring and apply it consistently across `check_property(property=...)`, `get_trace(...)`, `write_vcd(...)`, `disable_property(...)`, etc.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+    PROVED = "proved"
+    PROVED_WITH_BOUND = "proved_with_bound"
+    REFUTED = "refuted"
+    REFUTED_WITH_BOUND = "refuted_with_bound"
+    DISABLED = "disabled"
+    ASSUMED = "assumed"
+    UNSUPPORTED = "unsupported"
+    DROPPED = "dropped"
+    FAILURE = "failure"
+    INCONCLUSIVE = "inconclusive"
```
tautschnig, line 322: `PROVED` vs `PROVED_WITH_BOUND` and `REFUTED` vs `REFUTED_WITH_BOUND` directly mirror the internal C++ enum, but the naming is genuinely confusing to a first-time reader (the first sounds stronger, but it's the one with *no* bound info attached, not the one that proved at every bound). Consider:

- `PROVED_UNBOUNDED` (IC3/BDD, or k-induction that closed) vs `PROVED_UP_TO_BOUND` (BMC, with `Property.bound` populated), and symmetrically for refuted.

`FAILURE` vs `DROPPED` vs `UNSUPPORTED` vs `INCONCLUSIVE` is also a lot of distinctions — please document what each actually means, because they all look like "we gave up" to a Python user.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+class ElaborationError(EbmcError):
+    """Raised when elaboration fails."""
+
+class SolverError(EbmcError):
+    """Raised when the solver encounters an error."""
+
+class PropertyNotFoundError(EbmcError, KeyError):
+    """Raised when referencing a non-existent property."""
+```
+
```
tautschnig, line 391: The exception hierarchy is missing a `TypecheckError` (or similar). SystemVerilog typechecking happens inside `elaborate()`, and right now typecheck failures would have to be fused into `ElaborationError` (losing source location info) or `ParseError` (misleading). Also, only `ParseError` has structured `file`/`line`/`message` fields — all of the errors that originate from source code should carry that info.

Suggestion: define a small base mixin:

```python
class EbmcSourceError(EbmcError):
    file: str | None
    line: int | None
    message: str
```

and inherit `ParseError`, `TypecheckError`, `ElaborationError` from it. That gives tooling (e.g., an IDE wrapper, `pytest` reporter) one predictable way to surface source locations.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+design.read_verilog("dut.sv", systemverilog=True)
+design.set_top("dut")
+design.elaborate()
+
+traces = design.random_traces(count=100, steps=50, seed=42)
+for i, trace in enumerate(traces):
+    with open(f"trace_{i}.vcd", "w") as f:
+        f.write(trace.to_vcd())
+```
+
```
tautschnig, line 504: Worth calling out: what does the user see in the terminal / Python logs while `check_property(bound=50)` runs for minutes? The RFC mentions routing EBMC messages to `logging` further down, but doesn't say anything about:

- **GIL release** during the solver call. If the pybind11 layer holds the GIL for the duration of a solve, no other Python thread (UI, progress bar, logging handler with its own thread) can make progress.
- **SIGINT / Ctrl-C** handling. Users running this interactively will hit Ctrl-C. Does it cancel cleanly? Does the `Design` remain usable afterwards?
- **Progress / cancellation token**. For long checks, EDA users expect to be able to pass a timeout or a callback. Consider adding `timeout: float | None = None` and/or `on_progress: Callable[[Status], None] | None = None` on `check_property`.

These are all implementation details, but committing to them in the RFC avoids painful retrofitting later.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+# Build: G(counter != 0)
+counter = symbol("main.counter", type=unsigned_bv(8))
+zero = constant(0, type=unsigned_bv(8))
+prop_expr = G(counter != zero)
+
+design.add_property("my_prop", prop_expr)
+```
+
+This is a future extension; the initial release focuses on read-only
```
tautschnig, line 912: This is in direct tension with `Design.add_property_expr(name, expression: Expr)` at line 199 and the "Build: G(counter != 0)" example on lines 903-908. Either:

1. Expression construction is in the initial release, in which case `G`, `symbol`, `constant`, `unsigned_bv`, `__ne__` on `Expr`, etc. need real signatures in the RFC (especially: how are Verilog integer-promotion / width-matching rules handled for `counter != zero`?), or
2. Expression construction is explicitly deferred, in which case `Design.add_property_expr` should also be deferred — otherwise users have a method they can't call.

My preference is (2) for the initial release, documented as "deferred to v2". That gives time to design `Expr` construction properly (arguably the hardest API in the whole binding) without locking the shape in prematurely.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+    })
+    .def("__str__", [](const exprt &e) {
+        return format_to_string(e);
+    })
+    .def("__repr__", [](const exprt &e) {
+        return "<cprover.Expr id='" + id2string(e.id()) + "' operands=" +
+               std::to_string(e.operands().size()) + ">";
+    })
+    .def("__eq__", [](const exprt &a, const exprt &b) { return a == b; })
+    .def("__hash__", [](const exprt &e) { return e.hash(); });
```
tautschnig, line 948: A few issues with the illustrative pybind11 snippet that will bite during implementation (worth fixing now so the RFC doesn't mislead):

1. `e.find(dstringt(key))` — `irept::find` takes `const irep_idt &`, not `dstringt`. Use `e.find(key)` (implicit conversion) or `e.find(irep_idt{key})`.
2. `py::cast(id2string(val.id()))` — this assumes every named sub is itself an id-only irep. Many named subs hold a full expression or a string value under `ID_value`. `.get()` should probably return a richer type (`str | Expr | None`) or be split into `get_string()` / `get_expr()`.
3. `py::tuple(py::cast(e.operands()))` — `py::cast` of `std::vector<exprt>` returns a `py::list`, not something tuple-constructible directly. Use `py::cast(e.operands()).cast<py::tuple>()` or iterate explicitly.
4. `e.hash()` returns `std::size_t`; Python's `__hash__` wants a `Py_hash_t` (signed). Mask / truncate so negative-hash surprises don't show up on 32-bit platforms.

None of this changes the API shape — just marking the snippet as `// illustrative; see actual binding code` if you don't want to get into the weeds here.

doc/pyebmc-design.md
```diff
@@ -0,0 +1,1037 @@
...
+
+## Future Extensions
+
+- **Waveform viewer integration**: Return traces as pandas DataFrames.
+- **Async check**: `await design.check_property_async(...)` for long-running checks.
+- **Netlist access**: Expose gate-level netlist for custom analysis.
+- **Coverage**: `design.report_coverage()` after BMC runs.
+- **TCL compatibility shim**: A `pyebmc.tcl` module providing flat functions
+  (`read_verilog(...)`, `elaborate()`, etc.) operating on a global session,
+  for engineers migrating from Tcl scripts with minimal rewriting.
```
tautschnig, line 1037: **CAG (from @jimgrundy's review):** noted as requiring netlist manipulation, which is in "Future Extensions". Worth being explicit here about whether the netlist surface (read gates, insert observers, constrain nets) is something this RFC *commits to eventually exposing*, or whether it's out of scope. EDA users evaluating pyebmc will want to know before adopting it for CAG-style workflows.

Also missing from Future Extensions but worth having in the RFC so reviewers don't wonder: Python version support policy (the doc uses PEP 604 `str | Path` which is 3.10+, please state the minimum supported version), wheel/sdist distribution strategy (manylinux tags, macOS universal, Windows), and how the bundled solver licenses (MiniSat / CaDiCaL / Z3) compose with pyebmc's own license.

tautschnig (commented):
A few themes that I think need to be settled:

1. **Stringly-typed parameters** (`engine`, `solver`, `format`, property identifiers) should be enums or `typing.Literal` — both for IDE autocomplete and to keep the accepted set documented in one place.
2. **Property identity** is handled inconsistently (sometimes `name`, sometimes `identifier`, sometimes a `Property` — rarely all three). Pick one and let users pass `Property` objects everywhere an identifier is accepted.
3. **Expression construction** is listed as "future" but `add_property_expr` is in the initial API and `Design.add_constraint` takes strings — there's a mismatch that needs resolving before the RFC is accepted.
4. **Runtime concerns** that are invisible from a Python-level design but matter a lot for usability — GIL release during long solves, SIGINT handling, multiple concurrent `Design` instances, cleanup — aren't mentioned. An RFC is the right time to commit to them.

It would help to add a short **"Validation" section** outlining how the API will be exercised before release (doctest of the examples? an `examples/pyebmc/` regression target? unit tests for the binding layer?). That would also make it clear that the usage examples throughout are intended to stay runnable.

PR #1849, opened 2026-05-11
move rewrite for trivial `|=>` and `|->`

kroening:
This moves the trivial case of SVA `|=>` and `|->` where the LHS is not a proper sequence from `normalize_property` to `trivial_sva`.


tautschnig (approved):
Could we have a test for this, please?

kroening:
> Could we have a test for this, please?

added

PR #1866, opened 2026-05-23
Verilog: create symbol hierarchy directly

kroening:
This changes the way the Verilog typechecker processes submodule instances to create the final hierarchical instance identifiers directly.

This replaces the current approach where modules are type-checked for a given parameter assignment, and then subsequently instantiated by renaming the identifiers in the transition relation to match the instance identifier.

The key benefit is that the position in the module hierarchy is then known during type-checking, which enables upwards references.

### DIFF
--- a/src/verilog/verilog_synthesis.cpp
+++ b/src/verilog/verilog_synthesis.cpp
@@ -969,6 +969,7 @@ void verilog_synthesist::replace_by_wire(
   new_symbol.location=base.location;
   new_symbol.value=nil_exprt();
   new_symbol.is_auxiliary=true;
+  new_symbol.pretty_name = strip_verilog_root_prefix(new_symbol.name);
 
   symbol_exprt symbol_expr(new_symbol.name, new_symbol.type);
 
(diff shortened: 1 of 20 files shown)


src/verilog/verilog_synthesis.cpp
```diff
@@ -1352,15 +1353,16 @@ void verilog_synthesist::instantiate_port(
   const source_locationt &source_location,
   transt &trans)
 {
-  irep_idt port_identifier = port.identifier();
-
+#if 0
```
tautschnig: Let's either completely remove this block or else we need a comment explaining why we might want to re-enable it.

src/verilog/verilog_synthesis.cpp
```diff
@@ -1732,6 +1734,7 @@ void verilog_synthesist::expand_module_instance(
 
   replace_mapt replace_map;
 
+#if 0
```
tautschnig: As above: Let's either completely remove this block or else we need a comment explaining why we might want to re-enable it.

tautschnig (approved):

src/verilog/verilog_synthesis.cpp
kroening: removed

src/verilog/verilog_synthesis.cpp
kroening: removed

PR #1868, opened 2026-05-24
Verilog: factor out discovery of the symbols of an instantiated module

kroening:
This factors out the code that discovers the symbols of an instantiated module during synthesis.

### DIFF
--- a/src/verilog/verilog_synthesis.cpp
+++ b/src/verilog/verilog_synthesis.cpp
@@ -1685,6 +1685,34 @@ void verilog_synthesist::synth_module_instance_builtin(
 
 /*******************************************************************\
 
+Function: verilog_synthesist::find_module_symbols
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+std::vector<irep_idt>
+verilog_synthesist::find_module_symbols(const symbolt &module_symbol) const
+{
+  std::vector<irep_idt> result;
+
+  auto lower = symbol_table.symbol_module_map.lower_bound(module_symbol.module);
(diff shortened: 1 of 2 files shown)


src/verilog/verilog_synthesis.cpp
```diff
@@ -1685,6 +1685,34 @@ void verilog_synthesist::synth_module_instance_builtin(
...
+\*******************************************************************/
+
+std::vector<irep_idt>
+verilog_synthesist::find_symbols(const symbolt &module_symbol) const
+{
+  std::vector<irep_idt> result;
+
+  for(auto it =
+        symbol_table.symbol_module_map.lower_bound(module_symbol.module);
+      it != symbol_table.symbol_module_map.upper_bound(module_symbol.module);
```
tautschnig: Rather than relying on the compiler optimisation pass, shouldn't `symbol_table.symbol_module_map.upper_bound(module_symbol.module)` be stored in a temporary to make sure this is not recomputed upon each iteration?

src/verilog/verilog_synthesis.cpp
```diff
@@ -1685,6 +1685,34 @@ void verilog_synthesist::synth_module_instance_builtin(
...
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+std::vector<irep_idt>
+verilog_synthesist::find_symbols(const symbolt &module_symbol) const
```
tautschnig: Can we perhaps find a more descriptive name like `find_module_symbols` to avoid confusion with the global `find_symbols`? Also, wouldn't it be better to only pass a `const irep_idt &module`?

tautschnig (approved):

PR #1882, opened 2026-06-13
Verilog: based_unsized_literal_exprt

kroening:
Introduce `verilog_based_unsized_literal_exprt` for based unsized literals (e.g., `'hx0`, `'bx`, `'dz`).

Per IEEE 1800-2017 5.7.1, unsized literals where the MSB is x or z are extended to the size of the containing expression.

## Changes

- New `verilog_based_unsized_literal_exprt` class following the `verilog_unbased_unsized_literal_exprt` pattern
- `convert_literals` produces the new expression for based unsized literals (four-valued, two-valued, and `'dx`/`'dz` cases)
- `downwards_type_propagation` handles context-determined width via `with_context()`
- `verilog_lowering` expands to `constant_exprt` via `expand()`
- Replaces the previous `ID_C_verilog_unsized` attribute on `constant_exprt` for based literals

## Tests

- `integer_literals4`: promoted from KNOWNBUG to CORE (x/z extension to context width)
- `concatenation7`: new test for unbased unsized literal rejected in concatenation
- `assignment-to-part-select-of-struct1`: updated expectation for correct context-width extension


src/verilog/verilog_expr.cpp
```diff
@@ -454,6 +454,66 @@ verilog_unbased_unsized_literal_exprt::with_context(std::size_t width) const
...
+    {
+      // IEEE 1800-2017 5.7.1: context extension with x/z MSB
+      ext = bit_value[0];
+    }
+    else if(is_signed)
+      ext = bit_value.empty() ? '0' : bit_value[0];
+    else
+      ext = '0';
+    bit_value = std::string(width - bit_value.size(), ext) + bit_value;
+  }
```
tautschnig, line 493: I'd appreciate if the logic could be disentangled a bit:
1. Handle the case `bit_value.empty()` (which sets `bit_value` to `std::string(width, '0')`) upfront.
2. There are just two possible cases for `ext`: either `bit_value[0]` or `'0'`.

tautschnig (approved):

PR #1900, opened 2026-06-16
Verilog: error with location for $bits on module instance

kroening:
Previously, `$bits` applied to a module instance would produce a misleading error without source location (`type 'integer' has unknown number of bits`). Now it reports `cannot determine the number of bits` with the correct file and line.

The regression test `bits2.desc` is promoted from KNOWNBUG to CORE.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -1295,6 +1295,15 @@ exprt verilog_typecheck_exprt::convert_system_function(function_call_exprt expr)
         << base_name << " takes one argument";
     }
 
+    // Check that the argument type is valid for these functions.
+    if(
+      base_name == "$bits" &&
+      !verilog_bits_opt(arguments[0].type()).has_value())
+    {
+      throw errort().with_location(arguments[0].source_location())
+        << "cannot determine the number of bits of " << to_string(arguments[0]);
+    }
+
     // The return type is integer.
     expr.type() = integer_typet();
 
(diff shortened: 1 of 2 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -1295,6 +1295,15 @@ exprt verilog_typecheck_exprt::convert_system_function(function_call_exprt expr)
         << base_name << " takes one argument";
     }
 
+    // Check that the argument type is valid for these functions.
+    if(
+      base_name == "$bits" &&
```
tautschnig, line 1300: Is `$bits` really the only case here? E.g., is `$size` different in this regard?

kroening: Yes -- `$size` returns the number of elements in the dimension; you may well have an array of say class instances, and you can get the number of elements in the array, even if the class instances themselves don't have a number of bits. Similar for the others.

tautschnig: But then the situation is that for `$bits` we _know_ that something is going wrong here, but for others it might still go wrong? AFAIK one could use `$size` in place of `$bits` at times, for example.

tautschnig (approved):

src/verilog/verilog_typecheck_expr.cpp
kroening: The standard doesn't say what's right or wrong. Tool consensus appears to be to return 0. See #1903

PR #1927, opened 2026-06-21
fastppa: word-level delay, area, and power estimation tool

### LINKED ISSUE #9: ic3: property numbers
The 'short_name' idea seems to be that properties end on a dot followed by a unique number. This is not the case.

Instead of --prop, please use --property, to be consistent with the rest of ebmc.

kroening:
## Summary

New `fastppa` tool: pre-synthesis PPA estimation for Verilog/SystemVerilog at 3nm–130nm, ~**36–2550× faster** than logic synthesis (varies with design size/shape — see Speed below).

The numbers below are from real comparisons — Yosys 0.64 + ABC + OpenSTA (NanGate45 typical liberty, run in Docker for isolation/reproducibility) on 26 designs, plus a real Synopsys PrimeTime post-route report on an actual RISC-V core (Rocket, from the BSD-3-Clause [RTL-Timer](https://github.com/hkust-zhiyao/RTL-Timer) dataset). See `src/fastppa/DOCKER_VALIDATION.md` for full methodology and per-design results. An earlier version of this PR cited specific accuracy numbers against Synopsys PrimeTime; those were never actually backed by a script or license in this environment and have been replaced with these real, reproducible measurements. The comparison surfaced and fixed several real bugs in the cost model: FSM/case select logic costing zero, a miscalibrated register-area constant, an SRAM-density threshold that was far too low, struct/array-typed operands and registers being costed as 1 bit regardless of actual width, and an uncalibrated div/mod model that was 8× low on delay.

## Key Results

**26-design synthetic comparison** (Yosys+ABC+OpenSTA, NanGate45):

| Metric | Ratio (median / geomean / range) | Speed |
|--------|----------|-------|
| Area | 1.13× / 1.13× / 0.53–2.25× | 36×–2550× faster |
| Delay | 1.00× / 1.05× / 0.47–2.30× | than Yosys+ABC+OpenSTA |

Before any fixes, the same comparison gave area 1.58/1.66/0.26–7.50× and delay 0.63/0.52/0.05–1.63× on the original 21-design subset — i.e. up to 18× off on individual designs. The set grew to 26 designs in this update (4 pre-existing designs — `divider`, `parity`, `comb_mux`, `large_array` — had been omitted from the original comparison despite no cherry-picking being intended; `struct_mux` was added specifically to exercise the struct/array-width fix).

**Real-world check against commercial PrimeTime** (Rocket RV32 core, TinyRocket config, post-route NanGate45):

| | Area | Delay |
|---|---|---|
| fastppa vs PrimeTime (real) | **1.35×** | 1.95× |
| fastppa vs Yosys+ABC+OpenSTA (same RTL) | 2.16× | 1.10× |

fastppa's area is still closer to real commercial PPA than the open-source Yosys proxy used for the rest of the calibration (Yosys under-shoots at 0.62× of PrimeTime on this design). The delay gap to PrimeTime is mostly a single-pass-synthesis-vs-timing-closed-physical-design effect (fastppa tracks Yosys+ABC+OpenSTA on the same RTL at 1.10×) that only shows up on deep, real cores — not a calibration error in fastppa specifically.

Power was not independently validated (no real switching-activity ground truth available) and should be treated as a rougher estimate.

Known remaining limitations (found, documented, not yet fixed):
- Comparisons sharing the same operand pair (e.g. `<`, `==`, `>` on the same `a`,`b`) are costed independently; real synthesis often derives all three from one subtraction (confirmed ~2.2–3.0× over-estimate on the `cmp_*` calibration benchmarks).
- The same gap exists one level up for `div`/`mod`: a design computing both `a/b` and `a%b` is costed as two independent dividers, ~2.1× over, even though div/mod themselves are now individually well-calibrated (see Fixes below).
- No liveness analysis: a register with no observable output, or one that's fully written-then-read within a single cycle and never carries state across a clock edge, still gets charged full register cost — real synthesis dead-code-eliminates it. Confirmed on `counter` and (newly, after the struct/array-width fix stopped masking it) `struct_packet`.
- Array *write*-port decode logic (the `ID_with` array-update path) isn't costed by a decoder-shaped model at all — it falls through to a generic "simple logic" default bucket scaled by the *whole array's* aggregate width, not the element width. Confirmed via an ablation on `large_array` (drop the write port: the read side alone tracks real comb area closely). This is now the dominant source of area over-estimation on array-heavy designs (`regfile` 1.52×, `riscv_pipe` 1.69×) since the read-side fix below stopped a coincidental cancellation between an under-counted read and an over-counted write.
- fastppa's delay tracks single-pass logic synthesis, not post-timing-closure silicon; the gap to a fully optimized commercial flow grows with design depth (negligible on shallow designs, ~2× on a 39-level real RISC-V core).
- A real ASIC flow with a memory compiler would likely use an SRAM macro for a large register array; the open-source Yosys+ABC flow used for ground truth here has no memory compiler in the loop and just unrolls it into flip-flops. `large_array`'s area now happens to land at 1.02× (was 0.68×), but that's two errors canceling (an under-counted SRAM-density register model, and the over-counted array-write path above), not validated accuracy — see `DOCKER_VALIDATION.md` for the breakdown.
- A `case` statement with unsized integer literals as branch labels (`0` instead of `2'd0`) costs each branch's comparator at the Verilog-implied 32-bit comparison width rather than the width that actually matters, inflating cost ~6× — found while checking the FSM/case select-logic model, doesn't affect any existing regression design (they all use sized literals).

## Fixes in this PR (root-caused from real synthesis data)

1. FSM/case select logic was entirely uncosted (cost 0).
2. Register area constant was 2.2× too high vs real `DFF_X1`.
3. SRAM-density threshold was far too low (64 bits), miscosting ordinary register files.
4. Per-operator `+`/`-`/`*`/shift/comparator area & delay formulas recalibrated against real multi-width synthesis data.
5. FSM resource-sharing discount for iterative datapaths that reuse functional units across states.
6. Associative-chain delay balancing (`((a+b)+c)+d` costed as a balanced tree, not a literal chain).
7. Struct/array/packed-union-typed operands and registers were costed as 1 bit regardless of actual width (e.g. a struct-typed mux, an array-indexed register file) — now uses the Verilog frontend's own `verilog_bits_opt()`.
8. `div`/`mod` calibrated against real synthesis (was an unvalidated linear-in-width guess) — closes a delay gap that was 8× low.
9. Array read decode (`index_cost`) recalibrated for both delay and area, root-causing `regfile`'s long-standing 0.21× delay gap: the model only accounted for decode *depth* (`log2(N)` levels), missing that the address bits fan out to `~N` places and that fan-out's own delay dominates at `N >= 16`. Closes `regfile`'s delay gap (0.21× → 1.21×) and `large_array`'s (0.03× → 0.92×, 33× low to near-exact).
10. Constant array indices (`mem[3]`, or `mem[i]` after a bounded loop unrolls) were costed as if they needed a full decoder/mux tree; real synthesis just wires them directly, at zero cost. Found while validating fix #9 made one design (`array_sorter`, all-constant-index reads) worse, not better.

See `src/fastppa/DOCKER_VALIDATION.md` for the full root-cause writeup of each.

## Features

- Process nodes: 3nm 7nm 14nm 28nm 45nm 65nm 90nm 130nm
- `--clock-freq`, `--toggle-rate`, `--netlist`, `--process`
- RTL-derived activity ratio (auto clock-gating detection)
- Wire delay model (√area × fanout)
- FSM identity-branch optimization, FSM select/decode costing with cross-register sharing
- Constant-mult detection, SRAM density (above a 4096-bit threshold), operator merging
- Associative-chain delay balancing (matches real synthesis tree restructuring)
- Combinational design support

## Validated Against

- Yosys 0.64 + ABC + OpenSTA, NanGate45 typical liberty, run via the `openroad/orfs` Docker image — real synthesis and static timing, not estimated.
- Per-operator multi-width calibration benchmarks (`benchmarking/fastppa_scaling/`) used to fit the area/delay formulas for `+`, `-`, `*`, shifts, comparators, and div/mod against real synthesis data.
- A real commercial Synopsys PrimeTime post-route report (`benchmarking/rocket_core/`) on an actual RISC-V core, sourced from the BSD-3-Clause RTL-Timer dataset.

## Code Structure (~1700 lines C++)

```
src/fastppa/
├── main.cpp, fastppa_parse_options.{h,cpp}  — CLI
├── fastppa_frontend.{h,cpp}                 — Verilog → transition system
├── fastppa_estimator.{h,cpp}                — orchestration + output
├── estimate.{h,cpp}                         — core DAG walker
├── technology_db.{h,cpp}                    — process cost tables
├── README.md, BENCHMARKS.md
├── COMPARISON.md, CALIBRATION.md            — earlier, unverified numbers (kept for history)
├── DOCKER_VALIDATION.md                     — real methodology + results (see this one)
benchmarking/
├── fastppa_scaling/                         — per-operator, multi-width calibration designs
├── rocket_core/                             — real RISC-V core + PrimeTime ground truth
```

## Tests

- 59 regression tests (cli, blocks, designs incl. FPU/BF16, bugs), goldens regenerated against the fixed model
- Unit tests (179 assertions in 38 test cases)
- Oracle checks on exact area/delay/power values
- CI job in pull-request-checks.yaml


regression/fastppa/synth-fail-examples/README.md
```diff
@@ -0,0 +1,51 @@
+Synthesis Failure Analysis (TinyRocket, 263 modules)
```
tautschnig: Do we actually need this file?

src/fastppa/DOCKER_VALIDATION.md
```diff
@@ -0,0 +1,477 @@
+Docker-Based Validation: fastppa vs Real Synthesis
```
tautschnig, line 1: This file might be useful, but probably not all of its current contents.

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    if(e.id() != ID_constant)
+      return false;
+    auto val = id2string(e.get(ID_value));
+    if(val.empty() || val.back() != '1')
+      return false;
+    for(std::size_t i = 0; i + 1 < val.size(); i++)
+      if(val[i] != '0')
+        return false;
+    return true;
+  };
```
tautschnig: I think we have pre-existing utility functions for these.

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    {
+      op_cost.area_um2 *= OPERATOR_MERGE_FACTOR;
+      op_cost.energy_fj *= OPERATOR_MERGE_FACTOR;
+    }
+  }
+}
+
+/// Check if a binary expression is an identity/absorbing operation
+/// (e.g. x+0, x*1, x*0, x&all_ones) that synthesizes for free.
+static bool is_identity_op(const exprt &expr)
```
tautschnig, line 248: Should callers just use `simplify_expr`?

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,1042 @@
...
+    if(
+      op.id() == ID_symbol &&
+      resolved_constants.count(to_symbol_expr(op).get_identifier()))
+      return true;
+  }
+  return false;
+}
+
+/// Resolve an expression to a constant through typecasts and invar defs.
+static const exprt *resolve_to_constant(
```
tautschnig: I think we could use std::optional instead of a pointer.

tautschnig (commented):
I think we'll need to iterate on the actual software architecture a bit: this is making a lot of use of pointers, and the general algorithm is a bit obscured. Constant folding and expression simplification is done ad-hoc.

src/fastppa/estimate.cpp
kroening: done

src/fastppa/estimate.cpp
kroening: done

regression/fastppa/synth-fail-examples/README.md
kroening: removed

benchmarking/rocket_core/primetime/fetch_reports.sh
```diff
@@ -0,0 +1,58 @@
...
+# per-module area extract) from the full per-cell report, which is too
+# large (1 MB) to check in.
+#
+# Usage: ./fetch_reports.sh
+# Requires: curl, shasum (or sha256sum).
+
+set -eu
+cd "$(dirname "$0")"
+
+COMMIT=206ff4078368c251d2fafaffcc648282c68316f1
```
tautschnig, line 14: Why this specific commit and what does it take to upgrade to any more recent one?

benchmarking/rocket_core/README.md
```diff
@@ -0,0 +1,52 @@
...
+its submodules ultimately originate from the BSD/Apache-2.0-licensed
+[rocket-chip](https://github.com/chipsalliance/rocket-chip) /
+[chipyard](https://github.com/ucb-bar/chipyard) projects.
+
+Why this design: unlike the small, hand-written designs in
+`regression/fastppa/`, this is a real, deep (39 levels of logic per
+PrimeTime), multi-thousand-line RISC-V core with genuine commercial EDA
+ground truth, letting `fastppa`'s estimates be checked against more than
+just an open-source synthesis proxy. See
+`src/fastppa/DOCKER_VALIDATION.md` ("Real-world validation: Rocket core")
```
tautschnig, line 24: This DOCKER_VALIDATION.md file is a bit of a red herring: it contains a lot of information, basically a worklog. I believe we instead need two to three docs: 1) An architecture document that spells out what this tool does and how it does it. 2) A description of current limitations - some will be fundamental, others will just need more work. 3) A per-experiment document that spells out what candidate improvements have been identified that would have a material impact on the precision for that experiment.

benchmarking/rocket_core/README.md
```diff
@@ -0,0 +1,52 @@
...
+  logic (`Rocket.qor.rpt`)
+- Per-module cell area and power: `Rocket.hier_cell_area.txt`
+
+`compare_primetime.py` parses these reports, runs fastppa on the whole
+core and on each hierarchical submodule, and prints the comparison
+tables shown in `src/fastppa/COMPARISON.md`.
+
+(`fastppa`'s own delay number is closer to what a single-pass open-source
+synthesis flow, Yosys+ABC+OpenSTA, produces on the same RTL than to this
+post-route, timing-closed PrimeTime number -- see DOCKER_VALIDATION.md for
```
tautschnig, line 51: See above: I think we need to be much more explicit about this.

benchmarking/rocket_core/rocket_full.v
```diff
@@ -0,0 +1,6355 @@
+// See LICENSE.SiFive for license details.
```
tautschnig, line 1: We don't have this file here.

regression/fastppa/bugs/shared_comparator_bug.45nm.desc
```diff
@@ -0,0 +1,13 @@
...
+// lt/eq/gt all compare the same operand pair (a,b). Real synthesis
+// (confirmed via Yosys+ABC on the equivalent cmp_*.sv calibration
+// benchmarks) derives all three from one shared subtraction, costing
+// ~1x a single comparator, not 3x. Currently each is costed
+// independently (Operators: 3, Comb. area: 364.8 um2). See
+// src/fastppa/DOCKER_VALIDATION.md "Known limitation found but not fixed".
+^  Operators:           1$
+^  Comb. area:          121\.6 um2$
+^EXIT=0$
+^SIGNAL=0$
```
tautschnig, line 13: What exactly is the failing part of this test? Is it that we get the combinatorial area wrong?

src/fastppa/benchmark_scaling.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# Validate operator cost scaling across bit-widths.
+# Generates parameterized designs and compares fastppa vs Yosys.
+# Requires: Docker with hdlc/ghdl:yosys image.
+
+SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
+FASTPPA=${SCRIPT_DIR}/fastppa
+TMPDIR=$(mktemp -d)
+trap "rm -rf $TMPDIR" EXIT
```
tautschnig, line 9: Why don't we use the same in [benchmarking/rocket_core/primetime/fetch_reports.sh](https://github.com/diffblue/hw-cbmc/pull/1927/changes#diff-a378462b1afeabfbf597354bf3c48ecae917b560d8232a7df05ea12a8e2d0561) (and could then avoid git-ignoring `downloads/`)?

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+#include <util/std_types.h>
+
+#include <verilog/verilog_bits.h>
+
+#include <cmath>
+#include <optional>
+#include <unordered_map>
+#include <unordered_set>
+
+// --- Named constants for synthesis heuristic factors ---
```
tautschnig, line 24: It would be great to share how those factors were learned so as to know how to learn more factors should the need arise.

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+  /// Wire symbols currently being resolved (cycle detection).
+  symbol_sett &resolving;
+  /// Wire symbols whose logic has already been costed by an earlier
+  /// walk (a shared wire is one piece of hardware, not one per use).
+  symbol_sett &visited_invar_symbols;
+  /// Wire definitions collected from the invariant constraint.
+  const symbol_expr_mapt &invar_defs;
+  /// Wire symbols that transitively resolve to constants.
+  const symbol_expr_mapt &resolved_constants;
+  const technology_dbt &tech;
```
tautschnig, line 70: This one does not have a comment explaining what it is good for.

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+
+/// Compute the total bit-width of a type, including arrays, structs and
+/// (packed) unions.
+static std::size_t get_width(const typet &type)
+{
+  auto bits_opt = verilog_bits_opt(type);
+  if(bits_opt.has_value())
+    return bits_opt->to_ulong();
+  return 1;
+}
```
tautschnig, line 81: The comment does not do justice to the implementation - the actual width computation is done by `verilog_bits_opt` - the function magically adds a fallback, and the questions the comment should answer are: 1) why is the fallback safe and 2) why is "1" the correct fallback-value (and not, say, 0)?

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+    for(const auto &op : invar.operands())
+      collect_invar_defs(op, defs);
+  }
+  else if(invar.id() == ID_equal)
+  {
+    auto &eq = to_equal_expr(invar);
+    if(eq.lhs().id() == ID_symbol)
+      defs[to_symbol_expr(eq.lhs()).get_identifier()] = eq.rhs();
+    else if(eq.rhs().id() == ID_symbol)
+      defs[to_symbol_expr(eq.rhs()).get_identifier()] = eq.lhs();
```
tautschnig, line 98:
```suggestion
      defs[to_symbol_expr(eq.lhs()).identifier()] = eq.rhs();
    else if(eq.rhs().id() == ID_symbol)
      defs[to_symbol_expr(eq.rhs()).identifier()] = eq.lhs();
```

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+  }
+}
```
tautschnig, line 100:
```suggestion
static symbol_expr_mapt collect_invar_defs(const exprt &invar)
{
   symbol_expr_mapt defs;
   for(auto it = dest.depth_begin(), itend = dest.depth_end();
       it != itend;) // No ++it
   {
     if(it->id() == ID_equal)
     {
        auto &eq = to_equal_expr(*it);
        if(eq.lhs().id() == ID_symbol)
          defs[to_symbol_expr(eq.lhs()).identifier()] = eq.rhs();
        else if(eq.rhs().id() == ID_symbol)
          defs[to_symbol_expr(eq.rhs()).identifier()] = eq.lhs();
    }
     
    if(it->id() != ID_and)
      it.next_sibling_or_parent();
  }
  return defs;
}
```

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+  for(const auto &[id, def] : invar_defs)
+  {
+    exprt cur = def;
+    symbol_sett visited;
+    visited.insert(id);
+    // Follow symbol chains through typecasts
+    while(true)
+    {
+      while(cur.id() == ID_typecast && cur.operands().size() == 1)
+        cur = cur.operands()[0];
```
tautschnig, line 117:
```suggestion
      cur = skip_typecast(cur);
```

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+  symbol_expr_mapt &resolved_constants)
+{
+  for(const auto &[id, def] : invar_defs)
+  {
+    exprt cur = def;
+    symbol_sett visited;
+    visited.insert(id);
+    // Follow symbol chains through typecasts
+    while(true)
+    {
```
tautschnig, line 115: Shouldn't we do a union-find here rather than this potentially-O(n^2) traversal?

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+      }
+      else
+        break;
+    }
+  }
+}
+
+/// A node that is a literal constant or a symbol pre-resolved to one.
+/// Deliberately narrower than resolve_to_constant below: no typecast
+/// stripping, no hop through a non-constant invariant definition.
```
tautschnig, line 143: Hmm, but resolve_invar_constants (which populates resolved_constants) has done typecast stripping already.

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+static void apply_optimizations(
+  operator_costt &op_cost,
+  const exprt &expr,
+  std::size_t width,
+  std::size_t max_child_depth,
+  bool cmp_against_constant,
+  const walk_contextt &ctx)
+{
+  // Helper: check if an operand is constant, also accepting a symbol
+  // whose invariant definition is directly a constant.
```
tautschnig, line 162: Why the limitations to "directly a constant" rather than the transitive constant?

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+                     operand_is_constant(expr.operands()[1]);
+    if(has_const)
+    {
+      op_cost.area_um2 *= INCREMENTER_FACTOR;
+      op_cost.energy_fj *= INCREMENTER_FACTOR;
+    }
+  }
+
+  // MUX-tree cost for case/if-chains.
+  if(expr.id() == ID_if && max_child_depth > 2)
```
tautschnig, line 203: This does not seem to handle `ID_case`.

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+  {
+    bool child_same_op = false;
+    for(const auto &op : expr.operands())
+    {
+      if(op.id() == expr.id())
+      {
+        child_same_op = true;
+        break;
+      }
+    }
```
tautschnig:
```suggestion
    const bool child_same_op = expr.id() == expr.op0().id() || expr.id() == expr.op1().id();
```

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+      expr.id() == ID_le || expr.id() == ID_gt || expr.id() == ID_ge) ||
+    expr.operands().size() != 2)
+    return false;
+
+  for(const auto &op : expr.operands())
+  {
+    if(is_constant_node(op, ctx))
+      return true;
+  }
+  return false;
```
tautschnig:
```suggestion
   return is_constant_node(expr.op0(), ctx) || is_constant_node(expr.op1(), ctx);
```

src/fastppa/estimate.cpp
```diff
@@ -0,0 +1,966 @@
...
+  if(cur.id() == ID_constant)
+    return cur;
+  if(cur.id() == ID_symbol)
+  {
+    auto &sym_id = to_symbol_expr(cur).get_identifier();
+    auto rc = ctx.resolved_constants.find(sym_id);
+    if(rc != ctx.resolved_constants.end())
+      return rc->second;
+    auto it = ctx.invar_defs.find(sym_id);
+    if(it != ctx.invar_defs.end())
```
tautschnig, line 327: I'm confused: `ctx.resolved_constants` is (exclusively) populated by `resolve_invar_constants`, which in turn gets all its information from `ctx.invar_defs`. How is `resolve_to_constant` adding anything new here?

src/fastppa/estimate.cpp
kroening: done

src/fastppa/estimate.cpp
kroening: done

PR #1934, opened 2026-06-27
BTOR2: validate operand types and slice bounds

kroening:
Add input validation to the BTOR2 front-end to prevent invariant violations (aborts) on malformed inputs.

## Changes

1. **Array type validation**: Reject bitvector-only operators (`sext`, `uext`, `slice`, `redand`, `redor`, shifts, arithmetic, `concat`, overflow) when given array-typed operands. Previously these caused a precondition failure in `to_bitvector_type` or `boolbv_width`.

2. **Slice bounds checking**: Validate that slice upper/lower indices are within the source bitvector width. Previously out-of-bounds indices passed through to the solver, causing an invariant violation in `convert_extractbits`.

Both now produce a clean error message with exit code 6 instead of aborting.

## Testing

- 3 new regression tests added: `sext-array1`, `redand-array1`, `slice-oob1`
- All existing BTOR2 regression tests pass
- Full `regression/ebmc` test suite passes

## How bugs were found

Fuzz testing with [FuzzBtor2](https://github.com/CoriolisSP/FuzzBtor2) combined with manually crafted edge-case BTOR2 inputs targeting type mismatches and boundary conditions.

### DIFF
--- a/src/btor/btor_ebmc_language.cpp
+++ b/src/btor/btor_ebmc_language.cpp
@@ -172,6 +172,13 @@ static exprt build_op_expr(
   auto arg = [&](std::size_t i) -> exprt
   { return node_to_expr(node.args.at(i), model, expr_map); };
 
+  auto require_bv = [&](const exprt &e)
+  {
+    if(e.type().id() == ID_array)
+      throw ebmc_errort{} << "BTOR2: operator '" << node.tag
+                          << "' requires bitvector operands";
+  };
+
   const auto &tag = node.tag;
 
   // Unary operators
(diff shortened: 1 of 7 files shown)


src/btor/btor_ebmc_language.cpp
```diff
@@ -208,6 +232,8 @@ static exprt build_op_expr(
   // Signed/unsigned comparisons
   if(tag == "sgt" || tag == "sgte" || tag == "slt" || tag == "slte")
   {
+    require_bv(arg(0));
+    require_bv(arg(1));
```
tautschnig: Don't we almost-always require that all operands are bitvectors? Can we refactor the code so that we
1. First handle all operators that do _not_ require bitvectors.
2. Then iterate over all operands and do `require_bv`
3. Finally handle the remaining operators in the same way we did before?

kroening: Ok, done

tautschnig (approved):

PR #1935, opened 2026-06-27
AIGER: KNOWNBUG tests for front-end bugs found by fuzzing

kroening:
Added KNOWNBUG regression tests for three bugs in the AIGER front-end discovered using `aigfuzz` from Armin Biere's AIGER library:

1. **Undefined latch reset** (`undef_reset.aag`): Latch with `reset == lit` (non-deterministic initial value per AIGER spec) causes `unexpected AIGER reset value` error instead of being handled.

2. **Constraints ignored** (`constraint.aag`): AIGER constraints are parsed but never incorporated into the transition system, leading to spurious counterexamples.

3. **Duplicate symbol names crash** (`duplicate_names.aag`): Duplicate names (valid in AIGER) trigger an invariant violation in `build_transition_system()` at symbol table insertion.


src/aiger/aiger_ebmc_language.cpp
```diff
@@ -117,6 +117,27 @@ static transition_systemt build_transition_system(const struct aiger &model)
   // Map from AIGER variable index to symbol identifier
   std::map<unsigned, irep_idt> var_map;
 
+  // Helper to add a symbol, appending a numeric suffix on name collision.
+  auto add_symbol = [&](symbolt symbol) -> irep_idt
```
tautschnig, line 121: Can't we just use `symbol_table_baset::next_unused_suffix`?

kroening: done

tautschnig (approved):
Please amend the PR title and description.

kroening:
> Please amend the PR title and description.

Done

PR #939, opened 2025-01-17
Add chipsalliance/sv-tests to CI

kroening:
The chipsalliance/sv-tests are run on every merge to main, and the results are published to https://diffblue.github.io/hw-cbmc/

PR #1942, opened 2026-06-28
Add vlsim: Verilog simulator using the src/verilog front-end

kroening:
## Summary

- Adds `src/vlsim/` — a new binary `vlsim` that uses the existing `src/verilog` front-end to parse, preprocess, and elaborate Verilog/SystemVerilog files
- Supports `--preprocess`, `--show-parse`, and simulation (default) modes with the same CLI as `ebmc` for those operations
- Adds `examples/chipsalliance-sv-tests/vlsim.py` — an sv-tests runner for vlsim
- Adds `examples/chipsalliance-sv-tests/vlsim-sv-tests.sh` — a script to run the chipsalliance sv-tests suite against vlsim
- Adds a `sv-tests-vlsim` CI job that builds and runs the sv-tests suite with vlsim; the results are uploaded to sv-tests-vlsim/
- Updates `src/Makefile` to include vlsim in the default build


examples/chipsalliance-sv-tests/vlsim-sv-tests.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/sh
```
tautschnig, line 1: We need to de-duplicate this once #939 lands.

PR #1949, opened 2026-06-28
Verilog: handle deassign, release, and disable in convert_statement

kroening:
## Summary

`deassign`, `release`, and `disable` are valid procedural statements that reach `convert_statement` but were not handled, causing a `'unexpected statement'` error. They require no type-checking — add no-op cases for them.


tautschnig:
We should have one or more tests here.

kroening:
Now with a test for each of the statements.

tautschnig (approved):

PR #1957, opened 2026-07-01
KNOWNBUG test for comment in preprocessor macro

kroening:


regression/verilog/preprocessor/define-with-comment1.desc
```diff
@@ -0,0 +1,9 @@
+KNOWNBUG
+define-with-comment1.v
+--preprocess
+activate-multi-line-match
+^BAR1 \n  \n  BAR2$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
```
tautschnig, line 9: Would appreciate a description of what is going wrong here - neither does the commit message say anything to that effect, so it's hard to see.

kroening: added

regression/verilog/preprocessor/define-with-comment2.desc
```diff
@@ -0,0 +1,9 @@
+CORE
```
tautschnig, line 1: This one is failing in CI.

tautschnig (approved):

PR #1959, opened 2026-07-02
Verilog preprocessor: comments ending in backslash continue defines

kroening:
1800-2017 22.5.1: "A one-line comment (that is, a comment specified with the characters //) shall not be included in the substituted text. If a one-line comment ends in a backslash, the macro definition continues on the next line."

The tokenizer still excludes the comment from the macro text, but now returns the trailing backslash as a token while macro text is read, so that the `define continues on the next line.

This is on top of #1958 and #1957, and turns the KNOWNBUG test from #1957 into CORE.

### DIFF
--- a/regression/verilog/preprocessor/define-with-comment1.desc
+++ b/regression/verilog/preprocessor/define-with-comment1.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 define-with-comment1.v
 --preprocess
 activate-multi-line-match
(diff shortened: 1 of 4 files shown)


tautschnig:
Commit history needs a cleanup here (contains a merge commit).

kroening:
> Commit history needs a cleanup here (contains a merge commit).

Done

tautschnig (approved):

PR #1991, opened 2026-07-11
New IC3: report the length of the counterexample trace

kroening:
When the property is refuted, the IC3 solver now returns the length of the counterexample trace, counted in states: the trace has states at timeframes 0, ..., length − 1, with the property-violating state at timeframe length − 1.

- `ic3_resultt` is now a struct carrying the outcome and `counterexample_length`.
- Each proof obligation tracks its distance (in transitions) to the property-violating state; the frame level alone is not a reliable measure, since obligations are re-queued at higher levels.
- The trace is genuine, but not guaranteed to be a shortest counterexample: obligations are processed by frame level, not by depth (documented in the header).
- The refutation message now reads `Property refuted with counterexample of length N`; the `refuted1` regression test checks the reported length.
- Unit tests are adjusted to the new result struct; the refutation scenarios check the reported length, including a new deterministic 2-bit counter scenario whose only counterexample has four states.

### DIFF
--- a/regression/ebmc/new-ic3/refuted1.desc
+++ b/regression/ebmc/new-ic3/refuted1.desc
@@ -1,7 +1,7 @@
 CORE
 refuted1.sv
 --new-ic3 --top main
-^Property refuted$
+^Property refuted with counterexample of length 4$
 ^\[main\.p0\] always main\.cnt != 3: REFUTED$
 ^EXIT=10$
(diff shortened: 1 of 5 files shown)


tautschnig:
This needs git conflict resolution.

kroening:
done

tautschnig (approved):

PR #1993, opened 2026-07-12
New IC3: decision-variable restriction and eager clause pushing

kroening:
Two performance optimizations for the `--new-ic3` engine, motivated by profiling on the HWMCC08 sample from src/new-ic3/PERFORMANCE.md.

1. **Decision-variable restriction**: the per-frame MiniSAT solvers only branch on latch and input variables. The AIG gates use the full 3-clause Tseitin encoding, so BCP totalizes any assignment of latches and inputs; models remain genuine. Removes the `pickBranchLit` overhead (previously ~half of SAT search time); uniform 7–22% speedup.
2. **Eager clause pushing**: after MIC generalization, the blocking phase pushes the generalized clause to the highest frame where it is still relatively inductive — one SAT query per level — instead of re-running a full MIC generalization each time the obligation is re-queued one level higher. This fixes the engine's main weakness, refutations with deep counterexamples.

Results on the 49-benchmark HWMCC08 sample (60 s timeout, M2 Max; details added to PERFORMANCE.md):

| | solved (of 49) | timeouts | total time on solved |
|---|---:|---:|---:|
| before | 43 | 6 | 108.3 s |
| after | 45 | 4 | 88.9 s |

139453p22 (t/o → 5.5 s) and bc57sensorsp2 (t/o → 44.6 s) are newly solved; on the 43 benchmarks solved before, total time drops 108.3 s → 38.8 s (2.8x). Largest single improvements: neclaftp1001 16.0 → 1.3 s, nusmvtcasp5 24.3 → 4.7 s, 139462p24 13.9 → 0.8 s. Only regression: csmacdp2 1.3 → 4.7 s. prodcellp2 now terminates (~110 s) instead of diverging. All verdicts match the published HWMCC08 results, and the `regression/ebmc/new-ic3` tests pass.

A third experiment (VSIDS-style decay of the MIC literal-activity ordering) improved eijkbs3330 but caused a reproducible 15x regression on 139453p22 and was dropped.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1994, opened 2026-07-12
new IC3: handle invariant constraints and SVA assumptions

kroening:
The `--new-ic3` engine produces **wrong REFUTED verdicts** on transition systems with invariant constraints (SMV `INVAR`, AIGER 1.9 constraint sections). Found by evaluating on the HWMCC'25 bit-level safety benchmarks: 15 of the 142 sampled instances were refuted although the competition consensus (and BMC) shows them unsat — all 15 carry constraints. HWMCC'08/'17 and the Verilog regressions have no constraints, which is why this never surfaced. The bug is independent of #1993 (reproduces with the engine as on `main`).

**Root cause.** The constraints were asserted as unit clauses at the current frame in every solver. In the lifting solver, these units weaken `lift()`'s implication *cube ∧ inputs ∧ T ⇒ target* to hold only modulo the constraints, so a widened predecessor cube may include constraint-violating states; asserting such a cube through the next-state literals then admits transitions through constraint-violating states, and obligation chains built from them are spurious.

**Fix.** Keep the constraint units out of the CNF the lifting solver replays, and instead add the negated constraint conjunction to every lift target clause: the unsatisfiable core then proves *cube ∧ inputs ⊨ target ∧ constraints* unconditionally, so no widened cube contains a constraint-violating state. The property is weakened so a bad state must satisfy the constraints, and the constraint conjunction is asserted as a unit in the frame and init solvers (sound pruning). Every obligation chain is then a genuine constraint-satisfying counterexample: the initial state satisfies the constraints via the init solver, each lifted cube via its lift core, each unlifted (full-state) cube via the constraint unit in the frame solver that produced it, and the bad state via the weakened property. No extra state bits and no additional CNF are introduced; the cost is one literal per lift target.

(An earlier revision of this PR folded the constraints into a virtual `ok` latch instead; the lift-target formulation solves the same problem with less machinery and measures slightly better — 36 vs 35 solved on the 81 constrained instances of the HWMCC'25 sample.)

**Reproducer** (included as a regression test): a 3-latch system where the INVAR makes the bad state unreachable; `--new-ic3` reports REFUTED before the fix, while `--ic3` and BMC prove it. After the fix it is PROVED.

**Validation.**
- All 81 constrained instances in the HWMCC'25 sample: no wrong verdicts (before the fix: 15), 36 solved within 60 s.
- Unconstrained inputs take the identical code path (all guards are no-ops when there are no constraints); HWMCC'08 spot-checks unchanged.
- `regression/ebmc/new-ic3` passes (6 tests, including the new one).

### DIFF
--- a/src/new-ic3/ic3_solver.cpp
+++ b/src/new-ic3/ic3_solver.cpp
@@ -129,6 +129,16 @@ ic3_solvert::ic3_solvert(
   // Encode the netlist into CNF: one timeframe only — the next-state
   // functions are nodes in the same variable space.
   base_cnf = std::make_unique<recording_cnft>(solver_message_handler);
+
+  // Take the invariant constraints out of the netlist: unwind() would
+  // assert them as unit clauses in base_cnf, which every solver
+  // replays. In the lifting solver they would weaken lift()'s
+  // implication "cube ∧ inputs ∧ T ⇒ target" to hold only modulo the
+  // constraints, admitting constraint-violating states into widened
+  // cubes and hence spurious counterexample chains.
+  const auto constraints = std::move(netlist.constraints);
+  netlist.constraints.clear();
+
   bmc_mapt bmc_map(netlist, 1, *base_cnf);
   {
     messaget message{solver_message_handler};
@@ -172,12 +182,41 @@ ic3_solvert::ic3_solvert(
     }
   }
 
+  // The conjunction of the invariant constraints, as a literal at the
+  // current frame. It is asserted as a unit clause in the frame and
+  // init solvers, weakens the property (a bad state must satisfy the
+  // constraints), and is added to every lift target, so that the
+  // unsatisfiable core proves "cube ∧ inputs ⊨ target ∧ constraints"
+  // unconditionally: no widened cube can contain a
+  // constraint-violating state. Together this makes every obligation
+  // chain a genuine constraint-satisfying counterexample: the initial
+  // state satisfies the constraints via the init solver, each lifted
+  // cube via its lift core, each unlifted (full-state) cube via the
+  // constraint unit in the frame solver that produced it, and the bad
+  // state via the weakened property.
+  if(!constraints.empty())
+  {
+    literalt c_all = const_literal(true);
+    for(auto c : constraints)
+      c_all = base_cnf->land(c_all, bmc_map.translate(0, c));
+
+    if(!c_all.is_true())
+    {
+      constraint_lit = c_all;
+
+      // weaken the property: bad = c_all ∧ ¬P
+      prop_current = base_cnf->lor(!c_all, prop_current);
+    }
+  }
+
   lit_activity.resize(latches.size(), 0.0f);
 
   // Create the initial state solver.
   init_solver =
     std::make_unique<satcheck_no_simplifiert>(solver_message_handler);
   replay_base_cnf(*init_solver, true);
+  if(!constraint_lit.is_true())
+    init_solver->lcnf({constraint_lit});
 
   // Create lifting solver using IC3's MiniSAT (has releaseVar).
   lift_minisat = new_minisat_solver();
@@ -300,6 +339,8 @@ IctMinisat::Solver &ic3_solvert::get_solver(std::size_t level)
   if(!fs)
   {
     fs = new_minisat_solver();
+    if(!constraint_lit.is_true())
+      add_minisat_clause(*fs, {constraint_lit});
     if(level == 0)
       for(auto l : init_units)
         add_minisat_clause(*fs, {l});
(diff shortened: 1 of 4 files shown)


tautschnig:
How is this one related to #2013?

kroening:
> How is this one related to #2013?

Duplicate, closed #2013

yvizel:
A few comments:

** There's a redundant handling of the constraint when generating the BAD cube. This is probably not critical, but I am not sure if the expression is being simplified during the construction and gets rid of the extra redundant disjunct.

**Constant-false constraint conjunction** (`ic3_solver.cpp:204`). The guard is `!c_all.is_true()`, so if the constraints is actually `false`, probably better to report a vacuous result. 

**`netlist.transition` stays in the lifting solver.** Same dead-end reasoning applies to SMV `TRANS` if the relation is not total: a state in `s` with no successor under `i` satisfies `s ∧ i ∧ T ⇒ cube'` vacuously. 
Maybe this should be documented?

**Disabled assumptions** (`new_ic3_engine.cpp:86-99`). Assumptions are converted to constraints regardless of `is_disabled()`. If `--property` disables the others, do we still want a disabled assumption to constrain the check? Not sure if it matches BMC's behavior.

**Interface nit.** Taking `netlistt` by value and clearing `constraints` before `unwind` works, but a flag on `unwind` (or building `base_cnf` from a local copy behind a `const netlistt &`) would keep the mutation out of the public signature.

**Tests.** `constraint1.smv` is a nice minimal reproducer. Two gaps I'd fill: a *refuted* case where the constraint involves an input (all constrained cases here are on latches, and the input assignment in the lifted cube is where this bug lived), and a refuted case with an SVA assumption (`assume1.sv` only proves). A constrained refutation whose reported counterexample length differs from the unconstrained one would also be a good example.

kroening:
I've fixed the "interface nit" by adding the flag to (both variants of) `unwind`.

kroening:
assumptions: yes, assumptions don't get disabled when using `--property`. That behaviour should be consistent across engines.

kroening:
Tests added.

kroening:
Constant-false constraint: seems a very special case; reporting it is unlikely to add value.

kroening:
Non-left total transition relation: how is this typically handled in IC3?

yvizel:
> Non-left total transition relation: how is this typically handled in IC3?

It is usually not handled. We can have a comment in the code, but not necessary.

yvizel:
All good

src/new-ic3/ic3_solver.cpp
yvizel, line 1: In the comment, "not to asser" -- typo

src/trans-word-level/unwind.h
yvizel, line 1: "initial strate" -- typo

yvizel (approved):

kroening:
Fixed the typos, added a comment about the left-total assumption.

src/new-ic3/ic3_solver.cpp
```diff
@@ -130,10 +130,20 @@ ic3_solvert::ic3_solvert(
...
+    // In the lifting solver they would weaken lift()'s
+    // implication "cube ∧ inputs ∧ T ⇒ target" to hold only modulo the
+    // constraints, admitting constraint-violating states into widened
+    // cubes and hence spurious counterexample chains.
+    // The transition relation T is assumed to the left-total, which
+    // avoids that a state in s with no successor under i satisfies
+    // s ∧ i ∧ T ⇒ cube' vacuously.
     messaget message{message_handler};
-    ::unwind(netlist, bmc_map, message, *base_cnf, false, 0);
+    ::unwind(netlist, bmc_map, message, *base_cnf, false, false, 0);
```
tautschnig: Could we please do something along the lines of
```
unwind_optionst unwind_options;
unwind_options.add_initial_state = false;
unwind_options.add_constraints = false; // see comment above
::unwind(netlist, bmc_map, message, *base_cnf, unwind_options, 0);
```
with
```
struct unwind_optionst
{
  /// assert the initial-state constraint at timeframe 0
  bool add_initial_state = true;
  /// assert the in-state AIG constraints (netlist.constraints)
  bool add_constraints = true;
};
```
If and when we upgrade to C++ 20 we can then just do
```
::unwind(
  netlist, bmc_map, message, *base_cnf,
  {.add_initial_state = false, .add_constraints = false},
  0);
```
(GCC and Clang should support that also with earlier standards versions, but MSVC appears to be strict).

tautschnig (approved):

src/new-ic3/ic3_solver.cpp
kroening: adding the `unwind` interface changes as a separate commit

PR #1997, opened 2026-07-12
SMV: error on duplicate module names

kroening:
Duplicate `MODULE` declarations with the same name silently overwrote earlier entries in the module map, causing the analyzed module to potentially differ from the source.

This adds an error diagnostic that reports the duplicate at the second occurrence's source location. 

**Changes:**
- `src/smvlang/smv_typecheck.cpp`: Check for duplicate module names before building the index
- `regression/smv/modules/duplicate_module1.smv`: Test input with `MODULE main` declared twice
- `regression/smv/modules/duplicate_module1.desc`: Expects error with exit code 2

### DIFF
--- a/src/smvlang/smv_typecheck.cpp
+++ b/src/smvlang/smv_typecheck.cpp
@@ -28,6 +28,7 @@ Author: Daniel Kroening, kroening@kroening.com
 #include <cassert>
 #include <set>
 #include <stack>
+#include <unordered_set>
 
 class smv_typecheckt:public typecheckt
 {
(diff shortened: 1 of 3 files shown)


src/smvlang/smv_typecheck.cpp
```diff
@@ -3214,6 +3215,19 @@ void smv_typecheckt::typecheck()
...
+    std::unordered_set<irep_idt, irep_id_hash> module_names;
+    for(const auto &module : smv_parse_tree.module_list)
+    {
+      if(!module_names.insert(module.base_name).second)
+      {
+        throw errort().with_location(module.source_location)
+          << "duplicate module name `" << module.base_name << "'";
+      }
+    }
+  }
```
tautschnig: How about a `static void check_duplicate_module_names(const smv_parse_treet::module_listt &module_list)` which would avoid the need for a comment, the unnamed block, and avoid making `typecheck()`'s body even larger.

tautschnig (approved):

PR #2001, opened 2026-07-13
BDD: project inputs in EX

kroening:
## Summary
Project current input variables out of the BDD EX pre-image so CTL state sets do not depend on unconstrained inputs.

Promote EX_input1 from KNOWNBUG to CORE and add a unit test that exercises the input-projection behavior directly in bdd_model_checkert.

## Testing
- src/ebmc/ebmc regression/ebmc/BDD/EX_input1.smv --bdd
- src/ebmc/ebmc regression/ebmc/BDD/EX1.smv --bdd
- make -C unit

### DIFF
--- a/regression/ebmc/BDD/EX_input1.desc
+++ b/regression/ebmc/BDD/EX_input1.desc
@@ -1,4 +1,4 @@
-KNOWNBUG
+CORE
 EX_input1.smv
 --bdd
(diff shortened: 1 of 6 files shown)


tautschnig (approved):

tautschnig:
KNOWNBUG `next/next3.desc` has also been fixed it seems.

PR #2003, opened 2026-07-13
Temporal logic: restrict LTL_to_CTL to common fragment

kroening:
## Summary
Restrict LTL_to_CTL to the LTLdet fragment from Maidl's common-fragment result instead of recursively rewriting arbitrary F and G.

This keeps supported BDD LTL cases such as Gp and GFp, rejects unsupported formulas such as FGp and FXXp, and adds direct unit coverage for LTL_to_CTL.

## Testing
- make -C src/temporal-logic
- make -C src/ebmc
- make -C unit
- src/ebmc/ebmc regression/smv/LTL/smv_ltlspec1.smv --bdd
- src/ebmc/ebmc regression/smv/LTL/smv_ltlspec2.smv --bdd
- src/ebmc/ebmc regression/smv/LTL/smv_ltlspec_FG1.smv --bdd
- src/ebmc/ebmc regression/smv/LTL/smv_ltlspec_FX1.smv --bdd

### DIFF
--- a/unit/temporal-logic/ltl_to_ctl.cpp
+++ b/unit/temporal-logic/ltl_to_ctl.cpp
@@ -1,0 +1,124 @@
+/*******************************************************************\
+
+Module: LTL to CTL Unit Tests
+
+Author: Daniel Kroening, kroening@kroening.com
+
+\*******************************************************************/
+
+#include <temporal-logic/ctl.h>
+#include <temporal-logic/ltl.h>
+#include <temporal-logic/temporal_logic.h>
+#include <testing-utils/use_catch.h>
+
+SCENARIO("LTL_to_CTL accepts the supported fragment")
+{
+  auto p = symbol_exprt{"p", bool_typet{}};
+  auto q = symbol_exprt{"q", bool_typet{}};
+  auto r = symbol_exprt{"r", bool_typet{}};
+  auto s = symbol_exprt{"s", bool_typet{}};
+
+  THEN("Fp maps to AFp")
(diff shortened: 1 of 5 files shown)


unit/temporal-logic/ltl_to_ctl.cpp
```diff
@@ -82,23 +81,19 @@ SCENARIO("LTL_to_CTL accepts the supported fragment")
       *ctl ==
       AR_exprt{
         and_exprt{not_exprt{p}, r},
-        or_exprt{
-          and_exprt{p, AX_exprt{q}},
-          and_exprt{not_exprt{p}, r}}});
+        or_exprt{and_exprt{p, AX_exprt{q}}, and_exprt{not_exprt{p}, r}}});
```
tautschnig, line 84: Hmm, did clang-format changes get folded into this commit? I don't see a semantic difference in this line (similarly the case for others in this commit).

kroening: This file is entirely new -- there weren't any unit tests before.

tautschnig: Yes, but this change lives in the commit the commit message of which says "Temporal logic: avoid move-order bug in LTL_to_CTL" - which got me confused.

kroening: Squashed (the generated unit test had that move order bug)

tautschnig (approved):

PR #2010, opened 2026-07-16
hw-cbmc: make register_languages a proper override

kroening:
`hw_cbmc_languages.cpp` used to define the *base class's* `cbmc_parse_optionst::register_languages`, relying on the Makefile linking `cbmc_parse_options.o` while deliberately omitting `cbmc_languages.o`. This link-time translation-unit substitution breaks silently if the CBMC submodule ever restructures those translation units.

This PR overrides the virtual `register_languages` in `hw_cbmc_parse_optionst` instead and links the submodule's `cbmc_languages.o` normally, matching the pattern used by `ebmc_parse_optionst`. That requires additionally linking the `statement-list` and `json-symtab-language` libraries the base-class definition references (link-time only — those languages are not registered by hw-cbmc, since the override is what runs).

The class's remaining virtuals (`doit`, `help`) gain the `override` keyword to satisfy `-Winconsistent-missing-override`.

No functional change. hw-cbmc builds and the `regression/hw-cbmc` suite passes, including the Verilog tests that exercise the overridden language registration.

### DIFF
--- a/src/hw-cbmc/Makefile
+++ b/src/hw-cbmc/Makefile
@@ -4,6 +4,9 @@ SRC = hw_cbmc_main.cpp hw_cbmc_parse_options.cpp next_timeframe.cpp \
 
 OBJ+= $(CPROVER_DIR)/goto-checker/goto-checker$(LIBEXT) \
       $(CPROVER_DIR)/cbmc/cbmc_parse_options$(OBJEXT) \
+      $(CPROVER_DIR)/cbmc/cbmc_languages$(OBJEXT) \
+      $(CPROVER_DIR)/json-symtab-language/json-symtab-language$(LIBEXT) \
+      $(CPROVER_DIR)/statement-list/statement-list$(LIBEXT) \
       $(CPROVER_DIR)/goto-instrument/cover$(OBJEXT) \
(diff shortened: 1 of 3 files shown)


src/hw-cbmc/Makefile
```diff
@@ -4,6 +4,9 @@ SRC = hw_cbmc_main.cpp hw_cbmc_parse_options.cpp next_timeframe.cpp \
 
 OBJ+= $(CPROVER_DIR)/goto-checker/goto-checker$(LIBEXT) \
       $(CPROVER_DIR)/cbmc/cbmc_parse_options$(OBJEXT) \
+      $(CPROVER_DIR)/cbmc/cbmc_languages$(OBJEXT) \
+      $(CPROVER_DIR)/json-symtab-language/json-symtab-language$(LIBEXT) \
+      $(CPROVER_DIR)/statement-list/statement-list$(LIBEXT) \
```
tautschnig, line 9: Why is hw-cbmc linking in statement-list support?

kroening: This is bugging me as well, but doesn't have a good solution. The second you link cbmc's `register_languages` you need to link all the languages in there.

kroening: An option would be to use some trickery with static objects in those language modules.

PR #2019, opened 2026-07-16
new IC3: improved obligation priority and activity decay

kroening:
## Summary
- **Obligation priority**: use (level, depth, cube_size) ordering — lower level first, then deeper obligations (closer to counterexample root), then larger cubes (more specific, easier to block)
- **Activity decay**: multiply all literal activities by 0.95× at the start of each frame to keep MIC ordering adaptive

## Individual benefit
**5-10% fewer total queries** due to better work ordering. Deeper obligations are more likely to lead to actual counterexamples (worth exploring first). Activity decay prevents stale literals from dominating the drop order.

## Synergies
- **Parent-lemma** (PR #5): activity decay keeps recent parent literals relevant without artificial inflation
- **CTG/EXCTG** (PR #3): better obligation ordering means CTG encounters more productive cubes
- **Eager pushing** (PR #6): obligations re-queued at the pushed level benefit from depth tracking

## Test plan

### DIFF
--- a/src/new-ic3/ic3_solver.cpp
+++ b/src/new-ic3/ic3_solver.cpp
@@ -755,10 +755,17 @@ struct proof_obligationt
 {
   cubet cube;
   std::size_t level;
+  std::size_t depth;
 
   bool operator>(const proof_obligationt &other) const
   {
-    return level > other.level;
+    if(level != other.level)
+      return level > other.level;
+    // Prefer deeper obligations (closer to the counterexample root)
+    if(depth != other.depth)
+      return depth < other.depth;
+    // Prefer larger cubes (more specific states, easier to block)
+    return cube.size() < other.cube.size();
   }
 };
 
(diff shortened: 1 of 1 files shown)


src/new-ic3/ic3_solver.cpp
```diff
@@ -780,6 +787,10 @@ ic3_resultt ic3_solvert::solve()
     new_frame();
     std::size_t k = number_of_frames() - 1;
 
+    // Decay activity to keep the ordering adaptive
+    for(auto &a : lit_activity)
+      a *= 0.95f;
```
tautschnig, line 792: Why is the chosen magic number the right magic number?

tautschnig:
Should we look at using a different SAT solver instead?

kroening:
> Should we look at using a different SAT solver instead?

Yes, would seem appropriate. Will do the comparisons.

PR #2020, opened 2026-07-16
new IC3: recycle frame solvers after 2000 queries

kroening:
## Summary
- Track query count per frame solver
- After 2000 queries, rebuild the solver from base CNF + current frame clauses
- Adds `SOLVER_RECYCLE_LIMIT` constant and `frame_solver_queries` vector

## Individual benefit
**Prevents 2-5× gradual slowdown** in later frames. Frame solvers accumulate deactivated variables from `releaseVar` calls; after many queries the internal variable tables bloat, degrading unit propagation performance. Recycling restores a clean solver state.

## Synergies
- **Independent** — benefits all other optimizations by keeping frame solvers fast throughout the run
- Particularly important when combined with CTG (PR #3) and eager pushing (PR #6) which increase per-frame query counts

## Test plan

### DIFF
--- a/src/new-ic3/ic3_solver.cpp
+++ b/src/new-ic3/ic3_solver.cpp
@@ -295,11 +295,15 @@ std::unique_ptr<IctMinisat::Solver> ic3_solvert::new_minisat_solver()
 IctMinisat::Solver &ic3_solvert::get_solver(std::size_t level)
 {
   while(frame_solvers.size() <= level)
+  {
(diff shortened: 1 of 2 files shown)


tautschnig:
Why is 2000 the right number? And is that really the most appropriate solution? Is there no proper `reset` capability? I can't firmly say that this approach taken here is wrong, but it does not seem convincing.

PR #2027, opened 2026-07-19
Verilog: extend tests for `$typename`

kroening:
This adds further tests for the `$typename` system function.


regression/verilog/system-functions/typename1.desc
```diff
@@ -1,14 +0,0 @@
-CORE
```
tautschnig, line 1: The commit message only talks about additional tests so at most I was expecting some tests to have been moved, but this doesn't seem to be the case here?

kroening: This one got renamed to typename_bit1.desc -- git doesn't seem to track that when the files are short and something changes.

tautschnig: Ah, but then why were the real/shortreal/realtime cases removed?

kroening: Now in typename_real1, to make everything tidy.

tautschnig (approved):

PR #2028, opened 2026-07-19
SystemVerilog: `$typename` for logic types

kroening:
The `$typename` system function now correctly outputs the four-valued logic types.


tautschnig (approved):
This one now does need git conflict resolution (given my nit picking on the other PR... - sorry).

PR #2029, opened 2026-07-19
SystemVerilog: cast reset condition to Boolean

kroening:
If the reset condition given on the CLI is not Boolean, we now convert it.

### DIFF
--- a/src/verilog/verilog_ebmc_language.cpp
+++ b/src/verilog/verilog_ebmc_language.cpp
@@ -25,6 +25,7 @@ Author: Daniel Kroening, dkr@amazon.com
 #include "top_level_modules.h"
 #include "verilog_elaborate_compilation_unit.h"
 #include "verilog_language.h"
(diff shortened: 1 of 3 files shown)


src/verilog/verilog_ebmc_language.cpp
```diff
@@ -381,12 +382,17 @@ void verilog_ebmc_languaget::create_reset_logic(
     throw ebmc_errort{} << "failed to parse reset constraint";
   }
 
+  // must be boolean, and we lower it
+  auto reset_constraint_lowered =
+    verilog_lowering(typecast_exprt{reset_constraint, bool_typet{}});
```
tautschnig:
```suggestion
    verilog_lowering(typecast_exprt::conditional_cast(reset_constraint, bool_typet{}));
```
to avoid the unnecessary cast (in the typical case)?

tautschnig (approved):

PR #2044, opened 2026-07-22
logikbench.sh: cap per-circuit virtual memory

kroening:
## Summary
- The LogikBench CI job ([run 29873838090](https://github.com/diffblue/hw-cbmc/actions/runs/29873838090/job/88779864321)) died with exit code 143 ("runner has received a shutdown signal") right after `basic/bin2gray`.
- Root cause: the next circuit, `basic/bin2prio` (a 64-bit priority encoder with a `for` loop inside `always @(*)`), blows up ebmc's elaboration — reproduced locally at 16+ GB RSS within 8 seconds and still climbing. The existing 60s per-circuit `timeout` only guards against hangs, not memory exhaustion, so the GitHub-hosted runner (16 GB RAM) got OOM-killed before the timeout could fire, taking the whole job down with it.
- Fix: add a per-circuit `ulimit -v` cap (default 4 GiB) alongside the existing timeout, so such circuits fail fast and are recorded as a normal failure instead of exhausting the runner's memory. `ulimit -v` isn't available on macOS, so it degrades gracefully there (same behavior as before).

## Test plan
- [x] Verified `basic/bin2prio` reaches 16+ GB RSS locally without the cap.
- [x] Ran `logikbench.sh` locally against a small benchmark subset (`basic/band`, `basic/bin2gray`, `basic/bin2prio`); it now completes with exit 0 and correctly records the pathological circuit as a failure instead of hanging/OOMing.
- [x] `sh -n benchmarking/logikbench.sh` passes.
- [ ] Confirm the LogikBench GitHub Actions job goes green on Linux (where `ulimit -v` actually engages).

### DIFF
--- a/benchmarking/logikbench.sh
+++ b/benchmarking/logikbench.sh
@@ -101,8 +101,12 @@ for group in "$BENCHMARKS"/*/ ; do
     total=`expr $total + 1`
 
     # Parse and elaborate only (--bound 0).  Exit code 1 indicates a
-    # front-end/elaboration error; 124 is a timeout; anything else means ebmc
-    # processed the design (0 = holds/no properties, 10 = counterexample, ...).
+    # front-end/elaboration error; 124 is a timeout; 6 is CPROVER's
+    # catch-all for an internal exception, which is also what ebmc reports
+    # for std::bad_alloc (see CPROVER_EXIT_INTERNAL_OUT_OF_MEMORY in
+    # lib/cbmc/src/util/exit_codes.h) and thus what the PER_CIRCUIT_MEM_KB
+    # cap above triggers; anything else means ebmc processed the design
+    # (0 = holds/no properties, 10 = counterexample, ...).
     ( ulimit -v $PER_CIRCUIT_MEM_KB 2>/dev/null
       $RUN ebmc $incflags --bound 0 $sources ) > logikbench.out 2>&1
     status=$?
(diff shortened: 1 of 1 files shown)


benchmarking/logikbench.sh
```diff
@@ -96,7 +103,8 @@ for group in "$BENCHMARKS"/*/ ; do
     # Parse and elaborate only (--bound 0).  Exit code 1 indicates a
     # front-end/elaboration error; 124 is a timeout; anything else means ebmc
     # processed the design (0 = holds/no properties, 10 = counterexample, ...).
-    $RUN ebmc $incflags --bound 0 $sources > logikbench.out 2>&1
+    ( ulimit -v $PER_CIRCUIT_MEM_KB 2>/dev/null
+      $RUN ebmc $incflags --bound 0 $sources ) > logikbench.out 2>&1
```
tautschnig, line 111: The if/elif/else below will need to handle the out-of-memory exit code (presumably 143) as such rather than under the generic `FAIL` case.

kroening: Good catch. I tested this on Linux (built ebmc in an Ubuntu 24.04 container and reproduced `basic/bin2prio` under the same `ulimit -v` cap): ebmc actually catches `std::bad_alloc` itself and exits with status `6` (`CPROVER_EXIT_INTERNAL_OUT_OF_MEMORY` in `lib/cbmc/src/util/exit_codes.h`, printing `Out of memory`), not 143 — 143 was the *runner* being torn down by GitHub's infra before ebmc even hit the limit, which is exactly what this cap now prevents. Note that 6 is shared with `CPROVER_EXIT_EXCEPTION`/other internal-error codes, so it isn't purely an OOM signal, but it's ebmc's own catch-all for "something internal went wrong" as opposed to a front-end/elaboration error (1) or timeout (124). Added a dedicated `elif` branch (and report row styling) for status 6, labeled "out of memory / internal error", in c4e2ef87.

tautschnig (approved):

PR #2052, opened 2026-07-23
Verilog: fix three type-checker gaps from LogikBench triage

kroening:
This addresses the **type-checker gaps** bucket of the LogikBench RTL
triage (parse+elaborate only, `ebmc --bound 0`). Three independent
root causes, one commit each, each with a hand-authored regression test.

## Root cause 1 — "failed to get identifier on LHS constant" (genvar)
`parameterize_instantiated_modules()` did not restore the per-iteration
genvar values when descending into a `set_genvars` wrapper, so port
connections / parameter assignments of a generate-loop instance were
evaluated against the **post-loop** genvar value. A module output
connected to a genvar-indexed vector bit (`child u(.o(arr[g]))`) then
used the wrong index; for the top index `arr[WIDTH-1]` this looked
out-of-bounds, was folded to constant 0, and `check_lhs` rejected it.
Fix: restore (and afterwards re-restore) the captured genvar values,
mirroring `convert_module_item`.

- **blocks/lpddr5** — now fully passes.
- **blocks/chiplink**, **koios/bwave_like_fixed_small**,
  **koios/bwave_like_float_small** — clear this error; they now fail
  later on a *different, unrelated* synthesis issue ("conflicting
  assignment types" / "conflict with previous assignment"), which is a
  separate root-cause bucket, not fixed here.

## Root cause 2 — implicit nets in a concatenation LHS
IEEE 1800-2017 6.10 implicitly declares an undeclared identifier used on
the LHS of a continuous assignment as a net. This was handled for a bare
identifier but not for a concatenation member, e.g.
`assign {cout, out} = a - b;` (discard-carry idiom). Fix: declare each
undeclared bare-identifier member as a scalar net before converting.

- **arithmetic/sub** — now fully passes.

## Root cause 3 — `integer` has unknown number of bits
`verilog_bits_opt()` had no case for the mathematical integer type
`ID_integer` (produced by `$clog2` and unsized integer constants), so a
width query threw. Assigning such a value to a sized parameter reaches
`assignment_conversion`, which needs the width, e.g.
`localparam [2:0] ARSIZE = $clog2(DW/8);`. Fix: treat `integer` as 32
bits in width contexts, consistent with the existing
integer→`signedbv{32}` casts (IEEE 1800-2017 6.11, 11.8.1).

- **blocks/dma** — now fully passes.
- **koios/dnnweaver** — clears this error; then fails on an unrelated
  hierarchical reference into a generate-block array
  (`LOOP_N[n-1].buf_read_req_fwd`), a separate feature not in scope.

## Not fixed (left failing, with rationale)
- **large/vortex** — uses deep hierarchical references rooted at the
  module's own name into nested SystemVerilog interface instances
  (`vortex_core_wrap.core.warp_ctl_if.split_valid`). This is valid SV but
  needs substantial cross-module hierarchical-interface-reference support
  that EBMC does not have; out of scope for a targeted type-checker fix.
- **koios/dnnweaver** — beyond the `integer` fix it needs generate-block-
  array hierarchical references (see above).

## Summary of the 8 circuits
| Circuit | Status |
|---|---|
| arithmetic/sub | **fixed** |
| blocks/dma | **fixed** |
| blocks/lpddr5 | **fixed** |
| blocks/chiplink | past this error; blocked by unrelated synthesis issue |
| koios/bwave_like_fixed_small | past this error; blocked by unrelated synthesis issue |
| koios/bwave_like_float_small | past this error; blocked by unrelated synthesis issue |
| koios/dnnweaver | past this error; blocked by generate-block hierarchical ref |
| large/vortex | not fixed — needs hierarchical interface references |

`make -C regression/verilog test` passes with the three new tests added.

### DIFF
--- a/regression/verilog/continuous_assign_implicit_net1/main.v
+++ b/regression/verilog/continuous_assign_implicit_net1/main.v
@@ -1,0 +1,15 @@
+// An undeclared identifier used as a member of a concatenation on the LHS
+// of a continuous assignment is implicitly declared as a scalar net of the
+// default net type (IEEE 1800-2017 6.10). Here `cout` is never declared;
+// it is the discarded carry-out of the subtraction.
+// Regression for LogikBench arithmetic/sub.
+module main #(parameter DW = 16)
+  (
+   input [DW-1:0]  a,
+   input [DW-1:0]  b,
+   output [DW-1:0] out
+   );
+
+  assign {cout, out[DW-1:0]} = a[DW-1:0] - b[DW-1:0];
+
+endmodule
--- a/regression/verilog/continuous_assign_implicit_net1/test.desc
+++ b/regression/verilog/continuous_assign_implicit_net1/test.desc
@@ -1,0 +1,8 @@
+CORE
+main.v
+--module main --bound 0
+^Synthesis Verilog::main$
+^EXIT=10$
+^SIGNAL=0$
+--
+^.*unknown identifier cout.*$
--- a/regression/verilog/generate_instance_genvar1/main.v
+++ b/regression/verilog/generate_instance_genvar1/main.v
@@ -1,0 +1,21 @@
+// A module output port connected to a genvar-indexed bit of a vector
+// inside a generate for-loop. The genvar must be evaluated with its
+// per-iteration value when the port connection is elaborated; a regression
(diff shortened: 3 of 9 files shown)


kroening:
I'll split this PR up.

PR #2057, opened 2026-07-24
KNOWNBUG test for struct-typed parameters

kroening:

PR #2059, opened 2026-08-01
Verilog: accept struct/array/union as constant expressions

kroening:
Fixes the KNOWNBUG from #2057 by extending the constant expression check to accept struct, array, and union expressions with all-constant operands.

## Changes

- **`src/verilog/verilog_typecheck_expr.cpp`**: Added recursive `is_constant_rec()` helper and updated `elaborate_constant_expression_check()` to use it instead of the simple `is_constant()` check.
- **`src/verilog/verilog_simplifier.cpp`**: Added the same recursive check so the simplifier doesn't bail early on member access of struct-valued expressions (allowing `P.x` to be reduced to a constant).
- **Test**: Updated `struct_parameter1` from KNOWNBUG to CORE and added an assertion verifying the value.

## Testing

- `[main.assert.1] main.Q == 123: PROVED (tautology)`
- Full `regression/verilog` suite passes with no regressions.

### DIFF
--- a/src/verilog/verilog_typecheck_expr.cpp
+++ b/src/verilog/verilog_typecheck_expr.cpp
@@ -2210,6 +2210,36 @@ exprt verilog_typecheck_exprt::elaborate_constant_expression_rec(exprt expr)
 
 /*******************************************************************\
 
+Function: is_constant_rec
+
+  Inputs:
+
+ Outputs:
+
+ Purpose: returns true iff the given expression is a constant,
+          including struct, union and array expressions with
+          all-constant members
+
+\*******************************************************************/
+
+static bool is_constant_rec(const exprt &expr)
+{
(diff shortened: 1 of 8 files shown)


src/verilog/verilog_typecheck_expr.cpp
```diff
@@ -2210,6 +2210,36 @@ exprt verilog_typecheck_exprt::elaborate_constant_expression_rec(exprt expr)
...
+
+ Purpose: returns true iff the given expression is a constant,
+          including struct, union and array expressions with
+          all-constant members
+
+\*******************************************************************/
+
+static bool is_constant_rec(const exprt &expr)
+{
+  if(expr.is_constant() || expr.id() == ID_infinity)
```
tautschnig, line 2227: So we have two subtly-different implementations in two places - I think we need to consolidate.

PR #2063, opened 2026-08-05
extract `verilog_set_genvarst::build_map()`

kroening:
This extracts the code that turns the data in `verilog_set_genvarst` into a `map` into a method.


yvizel:
Earlier code had an exception thrown for the number of operands, now it is a precondition.
What happens if the precondition doesn't hold?

kroening:
> Earlier code had an exception thrown for the number of operands, now it is a precondition. What happens if the precondition doesn't hold?

The same; you get a runtime error.

tautschnig (approved):

PR #2072, opened 2026-08-09
New IC3: extract blocking phase

kroening:
Summary:
- extract the frame-blocking loop into a dedicated helper
- keep the top-level solve loop focused on frame creation, progress reporting, and propagation
- preserve the existing obligation-processing behavior while clarifying the phase split

Verification:
- make -C src/new-ic3 CXX='ccache clang++' CCACHE_DIR=/tmp/ccache-hw
- make -C src/ebmc CXX='ccache clang++' CCACHE_DIR=/tmp/ccache-hw
- ./src/ebmc/ebmc regression/ebmc/new-ic3/proved1.sv --new-ic3 --property main.p0
- ./src/ebmc/ebmc regression/ebmc/new-ic3/refuted1.sv --new-ic3 --top main
- ./src/ebmc/ebmc regression/ebmc/new-ic3/cover1.sv --new-ic3 --top main

### DIFF
--- a/src/new-ic3/ic3_solver.cpp
+++ b/src/new-ic3/ic3_solver.cpp
@@ -27,69 +27,91 @@ Author: Daniel Kroening, dkr@amazon.com
 #include <unordered_set>
 
 /// A cnf_clause_listt that can be instantiated (provides missing virtuals).
-class recording_cnft : public cnf_clause_listt {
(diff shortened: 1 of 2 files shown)


yvizel (approved):
Approved.

Small comment: the header says `block_frame(std::size_t level)` but the definition and call site use `k`.

PR #2081, opened 2026-08-11
Verilog: interface ports no longer depend on declaration order

kroening:
An ANSI port list could only use an interface port if the interface declaration had already been parsed. This made the order of the input files on the command line significant, so valid RTL was accepted or rejected depending purely on the file order.

## Example

```systemverilog
// mod.sv
module m(my_if.slave a, my_if.master b);
  assign b.valid = a.valid;
endmodule
```

```systemverilog
// iface.sv
interface my_if;
  logic valid;
  modport slave  (input valid);
  modport master (output valid);
endinterface
```

```
$ ebmc --systemverilog iface.sv mod.sv --show-modules   # accepted
$ ebmc --systemverilog mod.sv iface.sv --show-modules
file mod.sv line 1: syntax error, unexpected '.', expecting ')' or ',' before '.'
```

The same syntax error occurred for a single file in which the interface is declared after the module that uses it.

## Cause

The two grammar rules for interface ports in `ansi_port_declaration` were keyed on `TOK_INTERFACE_IDENTIFIER`. That token class is decided in the scanner (`make_identifier` in `src/verilog/scanner.l`, via `verilog_scopet::identifier_token`) and is only produced for identifiers that are already in the scope table as an interface. For an interface that has not been seen yet the scanner returns `TOK_NON_TYPE_IDENTIFIER`, and no port-list production matched.

## Fix

Interface ports are now recognised by their shape rather than by the token class: within a port list, an identifier followed by another identifier, optionally with an intervening `.modport`, can only be an interface port (IEEE 1800-2017 A.1.3, `interface_port_header`).

The two rules are moved into a new `interface_port_declaration` nonterminal, which is an alternative of `ansi_port_declaration_brace` rather than of `ansi_port_declaration`. That placement is what keeps the grammar LALR(1): the leading identifier is shifted before the nullable `attribute_instance_brace` has to be reduced. Adding the rules directly to `ansi_port_declaration` instead produces two shift/reduce conflicts. With this structure Bison reports no shift/reduce or reduce/reduce conflicts.

Both the `my_if.modport name` and the plain `my_if name` forms are fixed.

## Tests

New tests under `regression/verilog/interface/`:

- `port4` — modport form, with the interface file listed *after* the module file on the command line
- `port5` — no-modport form, same file ordering
- `port6` — modport form, interface declared after the module within a single file

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -11,6 +11,8 @@
 * Verilog: procedural assignments to hierarchical identifiers
 * SystemVerilog: ports for sequence and property declarations
 * SystemVerilog: $typename for logic types
+* SystemVerilog: interface ports no longer require the interface to be
+  declared before the module that uses it
 * Verilog: +define+ command-line option
 * Verilog: +incdir+ command-line option
 * Verilog: -l and +libfile+ command-line options
--- a/regression/verilog/interface/port4.desc
+++ b/regression/verilog/interface/port4.desc
@@ -1,0 +1,10 @@
+CORE
+port4B.sv
+port4A.sv --bound 5
+^\[main\.s\.p0\] always main\.s\.bus\.v == 1: PROVED up to bound 5$
+^EXIT=0$
+^SIGNAL=0$
+--
+--
+An interface port with a modport, where the interface is declared in a file
+that is given later on the command line.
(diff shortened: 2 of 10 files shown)


tautschnig (approved):
Approving, but needs git conflict resolution in the changelog.

kroening:
> Approving, but needs git conflict resolution in the changelog.

Done

PR #2107, opened 2026-08-18
Verilog: support macro default parameter values

kroening:
Implements default values for text macro formal arguments, e.g. ```define M(A, B=x)```, per IEEE 1800-2017 section 22.5.1.

## Background
The preprocessor previously rejected `=` in a macro parameter list with `default parameters are not supported yet` (added as KNOWNBUG in #2104).

## Change
- `definet::parametert` now carries an optional default token sequence.
- `parse_define_parameters` parses `= default_text`; the default text runs up to the next comma or the closing `)` at nesting depth zero, so commas nested inside `()`, `[]` or `{}` are part of the default.
- `parse_define_arguments` uses the default when an actual argument is omitted or empty (whitespace only). A value-less argument is an error only when the parameter has no default. The actual argument count may be fewer than the formal count (trailing defaults) but never more.

## Test
Promotes `regression/verilog/preprocessor/macro_default_param1` from KNOWNBUG to CORE and extends it:
```verilog
`define M(A, B=x) A+B          // M(a,b)->a+b   M(a)->a+x   M(a,)->a+x
`define N(A=1, B=2, C=3) A-B-C // N()->1-2-3 ... N(9,8,7)->9-8-7
`define P(A, B={1,2}) A-B      // P(p)->p-{1,2}  P(p,{3,4})->p-{3,4}
```

## Verification
- `make -C src/verilog` and `make -C src/ebmc` build cleanly.
- `make -C regression/verilog test`: all pass (226 skipped are KNOWNBUG/buechi), including the promoted `macro_default_param1` test.
- Manually verified: omitted arg without a default still errors; normal (non-default) macros unaffected.

## Blocked features
None.

### DIFF
--- a/src/verilog/verilog_preprocessor.cpp
+++ b/src/verilog/verilog_preprocessor.cpp
@@ -305,19 +305,69 @@ auto verilog_preprocessort::parse_define_parameters() -> definet::parameterst
     if(!parameter.is_identifier())
       throw verilog_preprocessor_errort() << "expecting a define parameter";
 
-    result.push_back(parameter.text);
+    definet::parametert define_parameter;
+    define_parameter.name = parameter.text;
 
     tokenizer().skip_ws();
 
     auto token = tokenizer().next_token();
 
+    if(token == '=') // SystemVerilog 2009: a default argument value
+    {
+      // 1800-2017 22.5.1: a formal argument may be followed by "= value",
+      // giving a default value that is used when the actual argument is
+      // omitted or empty. The default text runs up to the next comma or
+      // the closing parenthesis at nesting depth zero; commas nested
+      // inside (), [] or {} are part of the default text.
+      tokenizer().skip_ws();
+
+      std::vector<tokent> default_value;
+      std::size_t nesting = 0;
+
+      while(true)
+      {
+        if(tokenizer().eof())
+          throw verilog_preprocessor_errort()
+            << "eof inside a define parameter list";
+
+        if(
+          nesting == 0 &&
+          (tokenizer().peek() == ',' || tokenizer().peek() == ')'))
+        {
+          break;
+        }
+
+        auto value_token = tokenizer().next_token();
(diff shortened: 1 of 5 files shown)


src/verilog/verilog_preprocessor.cpp
```diff
@@ -305,19 +305,70 @@ auto verilog_preprocessort::parse_define_parameters() -> definet::parameterst
...
+        else if(
+          value_token == ')' || value_token == ']' || value_token == '}')
+          nesting--;
+
+        default_value.push_back(std::move(value_token));
+      }
+
+      // trim trailing whitespace of the default text
+      while(!default_value.empty() &&
+            (default_value.back() == ' ' || default_value.back() == '\t'))
```
tautschnig: Use isspace?

src/verilog/verilog_preprocessor.cpp
```diff
@@ -382,8 +433,20 @@ auto verilog_preprocessort::parse_define_arguments(const definet &define)
...
   }
 
-  // does the number of arguments match the number of parameters?
-  if(arguments.size() != define.parameters.size())
+  // An actual argument that is empty or whitespace only is treated as
+  // omitted (1800-2017 22.5.1).
+  auto is_empty = [](const std::vector<tokent> &tokens)
+  {
+    for(auto &token : tokens)
+      if(token != ' ' && token != '\t' && token != '\n')
```
tautschnig: What about \r?

tautschnig (approved):

src/verilog/verilog_preprocessor.cpp
kroening: Now uses helper method

src/verilog/verilog_preprocessor.cpp
kroening: now uses helper

PR #2108, opened 2026-08-18
Verilog: helper to detect white space

kroening:
This clarifies what is white space in the Verilog front-end: IEEE 1800-2017 5.3 lists "spaces, tabs, newlines, and formfeeds", which we interpret to be ' ', \t and \v, \r and \n, and \f.

`isspace` is not used since `isspace` requires the argument to fit into an unsigned char; however, our tokens use values beyond 255.

### DIFF
--- a/src/verilog/scanner.l
+++ b/src/verilog/scanner.l
@@ -23,6 +23,7 @@ static int isatty(int) { return 0; }
 
 #include "verilog_parser.h"
 #include "verilog_y.tab.h"
+#include "verilog_preprocessor_tokenizer.h"
 #include "verilog_scanner.h"
 
 int yyverilogerror(const char *error);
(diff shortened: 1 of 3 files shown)


src/verilog/scanner.l
```diff
@@ -34,17 +34,17 @@ static void line_directive()
   // skip directive
   while(*tptr!=' ' && *tptr!='\t' && *tptr!=0) tptr++;
 
-  // skip whitespace
-  while(*tptr==' ' || *tptr=='\t') tptr++;
+  // skip whitespace except newline
+  while(*tptr==' ' || *tptr=='\t' || *tptr=='\v' || *tptr == '\f') tptr++;
```
tautschnig: Shouldn't this use the new helper?

src/verilog/scanner.l
```diff
@@ -34,17 +34,17 @@ static void line_directive()
...
   PARSER.set_line_no(atoi(tptr)-1);
 
   // skip number
 
   while(isdigit(*tptr)) tptr++;
 
-  // skip whitespace
-  while(*tptr==' ' || *tptr=='\t') tptr++;
+  // skip whitespace except newline
+  while(*tptr==' ' || *tptr=='\t' || *tptr=='\v' || *tptr == '\f') tptr++;
```
tautschnig: Use the helper?

tautschnig (approved):

src/verilog/scanner.l
kroening: done

src/verilog/scanner.l
kroening: done

PR #2117, opened 2026-08-21
Verilog: register-transfer level IR layer

kroening:
This adds a new IR layer to the Verilog frontend that follows type checking and precedes synthesis.

## Summary

The register-transfer level (RTL) representation (`src/verilog/verilog_rtl.h`) is a map from identifiers to a map of slices (lower and higher bit index) to the definition of the slice, plus the assert, assume and cover properties of the module. The slice definition indicates whether the slice is state-holding (a register) or a wire, and gives the defining expression: the next-state value for state-holding slices, and the current-state value for wires.

The construction pass (`src/verilog/verilog_rtl.cpp`) builds the representation from the type-checked module items. It supports:

* clocked and combinational always constructs, continuous assignments, and net declarations with a value
* bit- and part-select lvalues, concatenation lvalues, members of packed structs and unions, array elements, and nestings thereof; constant selects map to slices, and array elements with non-constant indices yield a with-expression over the whole array; slices that are partially out of range are clipped per 1800-2017 11.5.1
* if statements (branch values are merged per bit-fragment at slice boundaries)
* case, casex and casez statements, converted into if-then-else chains with the same pattern masking for `?`, `x`, `z` as synthesis
* for loops, which are unrolled; break, continue and return use path-condition tracking, mirroring synthesis
* tasks and functions called as a statement, which are inlined; their locals are not part of the representation
* increment/decrement statements and compound assignment operators, lowered to plain assignments
* substitution of blocking assignments into subsequent right-hand sides, including reads after assignments to a part of a variable, which compose the value from the recorded fragments
* assert, assume and cover properties, recorded with their label when one is given: concurrent properties at module level, deferred assertions, and immediate and procedural concurrent assertions in always constructs (the path condition is applied as an implication)
* labeled statements

Constructs that synthesis also rejects (e.g. always_latch, guard-less always with assignments, procedural continuous assignments) produce a located error.

The new command-line option `--show-rtl` shows the representation, e.g.:
```
$root.main.q[3:0] register, next-state value: main.rst ? 0 : main.d[3:0]
$root.main.sum[7:0] wire, value: main.q + main.d
assert p1: main.q |-> main.d
cover c1: main.q == 15
```

## Full designs

The RTL representation now covers full designs, so that a transition system can be constructed from it (see #2119):

* Module instances are included recursively; port connections and primitive gates become constraints, and interface ports connect the members of the bound interface instance.
* Initial constructs and variable declarations with an initializer yield initial values, recorded per slice.
* The declared variables are recorded; identifiers forced by port connections are tracked so that they can become wires.
* User-defined function calls in right-hand sides are inlined, including output arguments and return statements; elaboration-time constants, hierarchical identifiers, named property/sequence instances, and constant system functions such as `$left` are resolved during substitution.
* Properties carry their type-checker identifier and context, and the default disable iff clause is applied per module and generate block.

## Testing

* New unit tests in `unit/verilog/verilog_rtl.cpp` (17 scenarios): slices, registers, wires, part selects, conditionals, case statements, for loops with and without break, task enables, concatenation lvalues, struct members, array elements, compound assignments, net declarations, increment/decrement, properties with and without labels, blocking-assignment substitution.
* Full unit suite passes (356 assertions in 62 test cases).
* `regression/verilog` and `regression/ebmc` suites pass.
* `--show-rtl` succeeds on all 746 tests in `regression/verilog` that reach RTL construction, except for 5 that deliberately provoke front-end errors.

### DIFF
--- a/src/verilog/verilog_rtl.cpp
+++ b/src/verilog/verilog_rtl.cpp
@@ -1,0 +1,3606 @@
+/*******************************************************************\
+
+Module: Verilog Register-Transfer Level Representation
+
+Author: Daniel Kroening, kroening@kroening.com
+
+\*******************************************************************/
+
+#include "verilog_rtl.h"
+
+#include <util/arith_tools.h>
+#include <util/bitvector_expr.h>
+#include <util/bitvector_types.h>
+#include <util/expr_util.h>
+#include <util/mathematical_types.h>
+#include <util/namespace.h>
+#include <util/simplify_expr.h>
+#include <util/std_expr.h>
+#include <util/symbol_table_base.h>
+
+#include <ebmc/ebmc_error.h>
+
+#include "aval_bval_encoding.h"
+#include "expr2verilog.h"
+#include "sva_expr.h"
+#include "verilog_bits.h"
+#include "verilog_expr.h"
+#include "verilog_typecheck_base.h"
+#include "verilog_typecheck_expr.h"
+#include "verilog_types.h"
+
+#include <algorithm>
+#include <ostream>
+#include <set>
+
+/*******************************************************************\
+
+   Class: verilog_rtl_buildert
+
+ Purpose: Constructs the RTL representation of a module from the
+          type-checked module items, before synthesis.
+
+\*******************************************************************/
+
+class verilog_rtl_buildert : public verilog_typecheck_baset
+{
+public:
+  verilog_rtl_buildert(
+    verilog_standardt _standard,
+    const symbol_table_baset &_symbol_table,
+    const namespacet &_ns,
+    const irep_idt &_module,
+    message_handlert &_message_handler)
+    : verilog_typecheck_baset(_standard, _ns, _message_handler),
+      module(_module),
+      symbol_table(_symbol_table),
+      message_handler(_message_handler)
+  {
+  }
+
+  void typecheck() override
+  {
+  }
+
+  // throws errort on error
+  verilog_rtlt build();
+
+protected:
+  const irep_idt module;
+  const symbol_table_baset &symbol_table;
+  message_handlert &message_handler;
+  verilog_rtlt rtl;
+
+  /// the context in which properties are recorded
+  verilog_rtl_propertyt::contextt property_context =
+    verilog_rtl_propertyt::contextt::MODULE_LEVEL;
+
+  /// the default disable iff condition, per module or generate block
+  std::optional<exprt> default_disable_iff;
+
+  using kindt = verilog_rtl_definitiont::kindt;
+
+  /// the state while processing the statements of one always construct
+  class statet
+  {
+  public:
+    /// values assigned so far, per identifier and slice
+    using slice_valuest = std::map<verilog_rtl_slicet, exprt>;
+    using value_mapt = std::map<irep_idt, slice_valuest>;
+
+    /// values assigned so far, by blocking and non-blocking assignments
+    value_mapt values;
+
+    /// values assigned by blocking assignments;
+    /// these are substituted into subsequent right-hand sides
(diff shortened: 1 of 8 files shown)


src/verilog/verilog_rtl.cpp
```diff
@@ -0,0 +1,3606 @@
...
+    /// and return statements to mark the rest of the path dead
+    exprt::operandst guard;
+  };
+
+  /// an lvalue, decomposed into its base symbol and the slice selected
+  class lhst
+  {
+  public:
+    symbol_exprt symbol;
+    verilog_rtl_slicet slice;
```
tautschnig, line 108: It would be helpful to have a comment explaining how the `slice` relates to `symbol` and what consistency invariants are required (I'd expect the slice having to be a valid interval with the symbol type's bit width, which in turn implies that the symbol's type has a defined bit width).

tautschnig (approved):

src/verilog/verilog_rtl.cpp
kroening: Comment added. Verilog explicitly allows out-of-bound slices, and hence, there are cases where the slice extends beyond the symbol's bit width.

src/verilog/verilog_rtl.cpp
```diff
@@ -0,0 +1,3615 @@
...
+
+  Inputs:
+
+ Outputs:
+
+ Purpose:
+
+\*******************************************************************/
+
+exprt verilog_rtl_buildert::read_lhs(exprt expr, statet &state)
```
yvizel, line 753: This function fetches the current value of a symbol with `fragment_value(..., whole_slice, ...)`. `fragment_value` only returns a recorded value when a *single* recorded slice covers the whole fragment; otherwise it falls back to the bare symbol. So a constant-slice assignment followed by a non-constant one on the same variable loses the first:

```systemverilog
reg [7:0] mem[0:15];
always @(posedge clk) begin
  mem[0] <= a;
  mem[i] <= b;
end
```

`--show-rtl` gives

```
$root.main.mem[127:0] register, next-state value: (main.mem WITH main.i:=main.b)
```

and `a` is gone. The same happens with aggregate struct members:

```systemverilog
typedef struct packed { logic [7:0] f; logic [3:0][7:0] arr; } s_t;
s_t s;
always @(posedge clk) begin
  s.f <= a;
  s.arr[i] <= b;
end
```

yields `s WITH arr := (s.arr WITH i := b)` with no trace of `a`. (The reverse order works, because `write_slice` splits the whole-symbol slice correctly.)

Maybe use `composed_value` instead?

src/verilog/verilog_rtl.cpp
```diff
@@ -0,0 +1,3615 @@
...
+    //   a[i]=e
+    // into
+    //   a = a WITH [i:=e]
+    auto &bit_select = to_verilog_bit_select_expr(lhs);
+    auto &src = bit_select.src();
+
+    auto old_value = read_lhs(src, state);
+    auto index = substitute(bit_select.index(), state);
+
+    with_exprt new_value{
```
yvizel, line 809: The `with` expression uses the raw Verilog index. `verilog_lowering` (and hence synthesis) maps Verilog indices to internal element positions (`(offset+size-1) - i` for descending unpacked ranges, `i - offset` for ascending ranges with an offset), and `decompose_lhs` applies that same mapping for constant indices. Within one RTL map the two paths disagree.
For `reg [7:0] mem[15:0]`:

- `mem[15] <= a` lands on slice `[7:0]` (internal element 0).
- `mem[i] <= b` becomes `mem WITH top.i:=top.b`, which for `i == 15` addresses internal element 15, i.e. bits `[127:120]`.

PR #2118, opened 2026-08-21
Verilog: fix upper bound when printing extractbits expressions

kroening:
The Verilog part-select rendered for an extractbits expression selects the bits from `index` to `index + width - 1`, but the printer emitted `index + width` as the upper bound. E.g., bits 4 to 7 of a vector were printed as `a[8:4]` instead of `a[7:4]`. The same off-by-one affected the non-constant index case, which now prints `a[i + 3:i]` instead of `a[i + 4:i]` for a 4-bit extraction.

## Testing

* New unit tests in `unit/verilog/expr2verilog.cpp` covering constant and non-constant indices.
* Full unit suite passes.
* `regression/verilog`, `regression/ebmc` and `regression/smv` suites pass.

### DIFF
--- a/src/verilog/expr2verilog.cpp
+++ b/src/verilog/expr2verilog.cpp
@@ -1122,13 +1122,13 @@ expr2verilogt::resultt expr2verilogt::convert_extractbits(
   if(src.index().is_constant())
   {
     auto index_int = numeric_cast_v<mp_integer>(to_constant_expr(src.index()));
-    dest += integer2string(index_int + width);
+    dest += integer2string(index_int + width - 1);
   }
   else
   {
     dest += convert_rec(src.index()).s;
     dest += " + ";
-    dest += std::to_string(width);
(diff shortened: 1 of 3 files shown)


src/verilog/expr2verilog.cpp
```diff
@@ -1122,13 +1122,13 @@ expr2verilogt::resultt expr2verilogt::convert_extractbits(
   if(src.index().is_constant())
   {
     auto index_int = numeric_cast_v<mp_integer>(to_constant_expr(src.index()));
-    dest += integer2string(index_int + width);
+    dest += integer2string(index_int + width - 1);
```
tautschnig, line 1125: Do we have a similar bug in CBMC?

tautschnig (approved):

src/verilog/expr2verilog.cpp
kroening: No, it's got the `-1`

PR #2124, opened 2026-09-06
Verilog: `bind` directive

### LINKED ISSUE #2102: Assertion binding not supported
There is a "unexpected TOK_NON_TYPE_IDENTIFIER" when using bind statement

See section 23.11 of IEEE STD 1800-2017

23.11 Binding auxiliary code to scopes or instances
It is often desired to keep verification code separate from the design code. SystemVerilog provides a bind
construct that is used to specify one or more instantiations of a module, interface, program, or checker
without modifying the code of the target. So, for example, instrumentation code or assertions that are
encapsulated in a module, interface, program, or checker can be instantiated in a target module or a module
instance in a non-intrusive manner. Similarly, instrumentation code that is encapsulated in an interface can
be bound to a target interface or interface instance.

kroening:
This adds support for the SystemVerilog `bind` directive, per 1800-2017 section 23.11. The directive specifies an instantiation of a module without modifying the source of the target module, which is used, e.g., to attach assertions or other instrumentation to a design in a non-intrusive manner.

## Changes
* Grammar: `bind_directive` (1800-2017 A.1.4) is now parsed, both at the top level and as a module item, producing a new `verilog_bind_directive` parse-tree node.
* Bind directives are handled during elaboration: compilation-unit level directives are registered when the compilation unit is elaborated, and directives that are module items are registered when the enclosing design element instance is elaborated. Registered directives are applied when a matching design element instance is elaborated, by adding the instantiation to the instance's copy of the module source, as if it was written there.
* Because registration happens during elaboration, a bind directive inside a generate construct only takes effect if its generate branch is taken.
* Both target forms are supported: a module target (`bind mod ...`) applies to every instance of the module, and an instance target (`bind top.sub ...`) applies to that one instance only. A directive that targets the design element instance currently being elaborated is applied in place.
* Errors are reported for directives that bind a module to itself, whose target does not exist, or whose target has already been elaborated (the target's item list is fixed at that point).
* Modules that are instantiated only through a bind directive are found when scanning library directories (`-y`).
* CHANGELOG entry.

## Testing
* New regression tests in `regression/verilog/bind/`: top-level bind (refuted bound assertion), bind as module item with a parameter assignment (proved), binds inside taken/untaken generate branches, an in-place bind targeting the enclosing module from a generate branch, an instance-target bind that verifies the sibling instance is unaffected, and error tests for unknown module targets, unknown instance targets, self-binds, and too-late binds.
* Full `regression/verilog`, `regression/ebmc`, `regression/vlindex` suites and the unit tests pass.

Fixes #2102


src/verilog/verilog_bind.cpp
```diff
@@ -0,0 +1,199 @@
...
+std::string bind_target_as_string(const exprt &target)
+{
+  if(target.id() == ID_hierarchical_identifier)
+  {
+    auto &binary = to_binary_expr(target);
+    return bind_target_as_string(binary.op0()) + '.' +
+           id2string(binary.op1().get(ID_base_name));
+  }
+  else
+    return id2string(target.get(ID_base_name));
```
tautschnig, line 34: In both of these uses of `.get(ID_base_name)`: how do we know that this `irept` entry exists (for what seemingly could be an arbitrary `irept`?

tautschnig (approved):

src/verilog/verilog_bind.cpp
kroening: Will fix.

kroening: Fixed in b68fdcc1: the target is now accessed via `to_verilog_identifier_expr` / `to_hierarchical_identifier_expr`, whose preconditions establish the expression kind (the grammar only produces these two forms for `bind_target`).

PR #2129, opened 2026-09-14
Verilog: upward name referencing

### LINKED ISSUE #2128: Upward name referencing
UPWARD hierarchical reference: a submodule reads a signal in a SIBLING instance by going up through the top-level module name. Standard SystemVerilog upward name referencing (IEEE 1800, "Upward name referencing"). VCS accepts it. EBMC reports "unknown identifier top".

module producer (clk);
  input  clk;
  wire   sig;
  assign sig = clk;
endmodule

module consumer (clk);
  input  clk;
  wire   probe;
  assign probe = top.prod.sig;   // <-- EBMC: unknown identifier 'top'
endmodule

module top ();
  wire clk;
  producer prod (.clk(clk));
  consumer cons (.clk(clk));
endmodule

kroening:
## Summary

Fixes #2128.

Per IEEE 1800-2017 §23.8, a hierarchical reference may start at the name of a
top-level module and descend into the instance hierarchy from there. This is
the mechanism by which a submodule reads a signal in a *sibling* instance:

```systemverilog
module producer(input clk);
  wire sig;
  assign sig = clk;
endmodule

module consumer(input clk);
  wire probe;
  assign probe = top.prod.sig; // <-- upward reference
endmodule

module top();
  wire clk;
  producer prod(.clk(clk));
  consumer cons(.clk(clk));
endmodule
```

`consumer` goes up through the top-level module name `top` to reach `prod.sig`.
VCS accepts this; EBMC previously rejected the leading name with
`unknown identifier top`.

The leading name of such a reference is a top-level instance registered under
`$root`. `convert_hierarchical_identifier` now recognizes a leading
`verilog_identifier` that names a top-level instance and rewrites it to that
instance symbol, from where the existing module-instance resolution descends
into the hierarchy (`.prod.sig`). The rewrite only applies when the name does
**not** resolve as an ordinary identifier in the current scope, so a local
object of the same name takes precedence (longest-prefix resolution, §23.8).

Based on #2040 (monolithic synthesis): that PR is what allows an assignment
crossing an instance boundary via a hierarchical identifier to be modeled
correctly, so upward references produce a sound transition system rather than
a silently over-constrained one.

## Test plan

- [x] `regression/verilog/hierarchical_identifiers/upward_name_referencing1`
  (new, CORE): `consumer` reads the sibling `producer`'s counter via
  `top.prod.cnt`; `p_ok` (probe mirrors the counter) is PROVED and `p_bad`
  (probe constantly zero) is REFUTED, confirming the reference is wired to the
  sibling's value.
- [x] `regression/verilog/modules/upwards_reference1` flipped from `KNOWNBUG`
  to `CORE`: pre-existing test for an upward reference to a top-level
  parameter (`top.P`), now type-checks and proves.
- [x] A local object named the same as a top-level module correctly shadows
  the module name (verified manually).
- [x] Full `regression/verilog` suite passes.
- [x] Full `regression/ebmc` suite passes.


kroening:
Depends on https://github.com/diffblue/hw-cbmc/pull/2040

